PR #9, opened 2025-07-01
Add initial implementation of getting device ID in runtime

jmasek:


zephyr/src/net/spotflow_device_id.c
```diff
@@ -0,0 +1,66 @@
...
+char spotflow_generated_device_id_buffer[(2 * ZEPHYR_DEVICE_ID_MAX_LENGTH) + 1];
+
+char* cached_device_id = NULL;
+
+char* __attribute__((weak)) spotflow_override_device_id()
+{
+	return NULL;
+}
+
+void spotflow_generate_device_id()
```
jmasek, line 21: should be static, please move after public (non-static) methods

zephyr/Kconfig
```diff
@@ -62,14 +63,16 @@ if LOG_BACKEND_SPOTFLOW
 config SPOTFLOW_DEVICE_ID
 	string "Device ID device will use to connect to Spotflow observability platform"
 	help
```
jmasek, line 65: missing default value or de we want users to send empty device id?

zephyr/src/net/spotflow_device_id.c
```diff
@@ -0,0 +1,66 @@
...
+{
+	if (strlen(spotflow_generated_device_id_buffer) > 0) {
```
jmasek, line 23: to discuss - you are expecting the buffer to be zeroed. Not sure if it is a good practice.

zephyr/src/net/spotflow_device_id.c
```diff
@@ -0,0 +1,66 @@
...
+
+#include "spotflow_device_id.h"
+
+/* Should be safe upper bound - used, for example, in Zephyr Shell */
+#define ZEPHYR_DEVICE_ID_MAX_LENGTH 16
+
+LOG_MODULE_REGISTER(spotflow_device_id, CONFIG_SPOTFLOW_PROCESSING_BACKEND_LOG_LEVEL);
+
+/* Hexadecimal representation of the Zephyr device ID and a null terminator */
+char spotflow_generated_device_id_buffer[(2 * ZEPHYR_DEVICE_ID_MAX_LENGTH) + 1];
```
jmasek, line 12: buffer is only used for generation. Does it make sense to have it in the module and not only in the generation logic?

zephyr/src/net/spotflow_device_id.c
```diff
@@ -0,0 +1,66 @@
...
+
+char* cached_device_id = NULL;
```
jmasek, line 14: static?

zephyr/src/net/spotflow_mqtt.c
```diff
@@ -188,12 +188,16 @@ static int client_init(struct mqtt_client* client)
 						      spotflow_mqtt_config.server_addr,
 						      spotflow_mqtt_config.port);
 
+	char* device_id = spotflow_get_device_id();
+	mqtt_client_toolset.username =
+	    (struct mqtt_utf8){ .utf8 = device_id, .size = strlen(device_id) };
```
jmasek, line 193: I do not understand this and why it is added to toolbox. 
Why are you not doing something simple like this
```
char* device_id = spotflow_get_device_id();
client->user_name =   (struct mqtt_utf8){ .utf8 = device_id, .size = strlen(device_id) };
```

roberthusak:
The comments will be addressed in https://github.com/spotflow-io/device-sdk/pull/11

zephyr/Kconfig
roberthusak: If empty, device SDK uses `hwinfo_get_device_id` - see the `help` text.

zephyr/src/net/spotflow_device_id.c
roberthusak: We can't have it as a local variable in `spotflow_generate_device_id`, because it stores the string to which a pointer will be returned (and returning from the function would invalidate it).

We can make it a static variable in the function, but I'm not sure if that's a good practice.

zephyr/src/net/spotflow_device_id.c
roberthusak: Fixed.

zephyr/src/net/spotflow_device_id.c
roberthusak: Fixed.

zephyr/src/net/spotflow_device_id.c
roberthusak: The logic is correct - the buffer **will** be zeroed as well as any other static variable. However, I added a boolean flag for better readability.

zephyr/src/net/spotflow_mqtt.c
roberthusak: `client->user_name` requires a pointer to `struct mqtt_utf8`. Therefore, I had to put it in a place in memory that will be later valid when used. I put it to `mqtt_client_toolset`, because `spotflow_mqtt_config` seem to contain only static data.

However, now I noticed that `server_addr` is initialized dynamically as well, so I put it back to `spotflow_mqtt_config`.

zephyr/src/net/spotflow_device_id.c
jmasek: I see

zephyr/src/net/spotflow_device_id.c
jmasek: actually, I find a bit confusing.
this should never happen, right? 
```
static void spotflow_generate_device_id()
{
	if (device_id_generated) {
		return;
	}
```

jmasek: ok, it can - if this
``` 
ssize_t ret = hwinfo_get_device_id(device_id, ARRAY_SIZE(device_id));

	if (ret <= 0) {
	```
	but it won't be retried anyway, because you set cached_device_id = device_id; so it cannot happen, I assume, right.

roberthusak: You're right, `device_id_generated` was a redundant layer of caching, so I removed it.

PR #11, opened 2025-07-03
Add initial implementation of getting device ID in runtime

roberthusak:
Re-created https://github.com/spotflow-io/device-sdk/pull/9 (responded to comments there)


zephyr/Kconfig
```diff
@@ -1,6 +1,7 @@
 menuconfig SPOTFLOW
 	bool "Enable Spotflow Observability"
 	default n
+	select HWINFO
```
jmasek, line 4: do we want to enable this everytime with spotflow? it is needed only if device id is not provided, right?

roberthusak: I tried to make the selection of `HWINFO` conditional, but it's AFAIK impossible to do it based just on the value of the string option `SPOTFLOW_DEVICE_ID` - `if` clauses require boolean or tri-state options. We could achieve this by exposing an additional option(s) to the user, but I wouldn't like to do it so that the interface is as simple as possible.

zephyr/src/net/spotflow_device_id.c
```diff
@@ -0,0 +1,71 @@
...
+
+LOG_MODULE_REGISTER(spotflow_device_id, CONFIG_SPOTFLOW_PROCESSING_BACKEND_LOG_LEVEL);
+
+/* Hexadecimal representation of the Zephyr device ID and a null terminator */
+static char spotflow_generated_device_id_buffer[(2 * ZEPHYR_DEVICE_ID_MAX_LENGTH) + 1];
+static bool device_id_generated = false;
+
+static char* cached_device_id = NULL;
+
+char* __attribute__((weak)) spotflow_override_device_id()
```
jmasek: Typically, I saw different order `__attribute__((weak)) char* spotflow_override_device_id()`

jmasek: const char*?

zephyr/src/net/spotflow_device_id.c
```diff
@@ -0,0 +1,71 @@
...
+{
+	return NULL;
+}
+
+static void spotflow_generate_device_id();
+
+char* spotflow_get_device_id()
```
jmasek: const?

zephyr/src/net/spotflow_device_id.c
```diff
@@ -0,0 +1,71 @@
...
+
+#include "spotflow_device_id.h"
+
+/* Should be safe upper bound - used, for example, in Zephyr Shell */
+#define ZEPHYR_DEVICE_ID_MAX_LENGTH 16
+
+LOG_MODULE_REGISTER(spotflow_device_id, CONFIG_SPOTFLOW_PROCESSING_BACKEND_LOG_LEVEL);
+
+/* Hexadecimal representation of the Zephyr device ID and a null terminator */
+static char spotflow_generated_device_id_buffer[(2 * ZEPHYR_DEVICE_ID_MAX_LENGTH) + 1];
```
jmasek, line 14: I would prefer `(2 * ZEPHYR_DEVICE_ID_MAX_LENGTH * sizeof(char)) + 1`

zephyr/src/net/spotflow_device_id.c
```diff
@@ -0,0 +1,71 @@
...
+	if (ret <= 0) {
+		LOG_ERR("Failed to get Zephyr device ID (%d), using default", (int)ret);
+
+		strncpy(spotflow_generated_device_id_buffer, "default_device_id",
+			sizeof(spotflow_generated_device_id_buffer));
+		return;
+	}
+
+	for (int i = 0; i < ret; i++) {
+		snprintk(spotflow_generated_device_id_buffer + (2 * i), 3, "%02X", device_id[i]);
```
jmasek, line 68: could you please add a comment to clarify whats going on?

jmasek (changes requested):

zephyr/src/net/spotflow_device_id.c
roberthusak: Done.

zephyr/src/net/spotflow_device_id.c
roberthusak: Done.

zephyr/src/net/spotflow_device_id.c
roberthusak: I saw it both ways, but yours seems more logical, fixed.

zephyr/src/net/spotflow_device_id.c
roberthusak: Not sure about that - the semantics of `[...]` is AFAIK to specify the length of the array (and this one is already of type `char`) while `sizeof(...)` tells the number of bytes. So, `some_type a[n * sizeof(some_type)]` would usually take up more bytes than necessary unless `sizeof(some_type)` is `1` as in the case of `char`.

I extended the comment. Is it sufficient now?

jmasek: ok

jmasek (dismissed):

jmasek (approved):

PR #14, opened 2025-07-08
added west-zephyr for direct init

jmasek:


jmasek:
Won't do it after the discussion with the team. This option would confuse the user. Preferred way is to use `west init --mr ` to have specific version of zephyr

PR #26, opened 2025-08-25
added eth sample

jmasek:


zephyr/samples/coredumps/src/main.c
```diff
@@ -19,6 +28,12 @@ static struct gpio_callback button_cb_data;
 
 static int prepare_button();
 
+#ifdef CONFIG_ETH_DRIVER
+static void turn_on_dhcp_when_device_is_up();
+#endif
+
+//to use nxp with ethernet, use build with -DOVERLAY_CONFIG="boards/frdm_rw612_eth.conf"y
```
roberthusak: Obsolete comment?

zephyr/samples/coredumps/Kconfig
```diff
@@ -8,4 +8,33 @@ config NET_WIFI_PASSWORD
 	help
 		Wifi password for the network that the device should connect to.
 
+config WIFI
+	bool
+	default y
+	imply WIFI_NM
+	imply NET_L2_WIFI_MGMT
+
+
```
roberthusak, line 39: Nitpick: Keep consistent spacing between options (2 empty lines here and no empty line below). Update it in the second sample as well.

zephyr/samples/coredumps/Kconfig
```diff
@@ -8,4 +8,33 @@ config NET_WIFI_PASSWORD
...
+	default y if WIFI
+	default n if !WIFI
+
+# Enables IP fragmentation, which fixes the TLS handshake issue in Zephyr 4.2
+config NXP_WIFI_TX_RX_ZERO_COPY
+	bool
+	default y if WIFI_NXP
+	default n if !WIFI_NXP
+
+# end of NXP specific configurations
```
roberthusak: Wouldn't be shorter and more reliable (i.e. not depending on the order how the config options are loaded) to extend `config WIFI` with the following? (possibly `imply` instead of `select` where it makes sense)
```
	select WIFI_NXP if NXP_RW610
	select NXP_WIFI_TX_RX_ZERO_COPY if NXP_RW610
```

I'm also thinking whether not to introduce something like `config SAMPLE_WIFI` and `config SAMPLE_ETH` (or `NET_WIFI` and `NET_ETH`?) and only `select`/`imply` the existing Zephyr config options within each of them. This approach would not only be resilient to the loading order of config options, but it could also better serve as a documentation - now it might confuse the developer that we're redefining options defined elsewhere.

But maybe this suggested approach is overengineered, I'll leave it up to you 😅

roberthusak (commented):
Looks good, just some minor changes and one idea.

zephyr/samples/coredumps/Kconfig
jmasek: fixed

zephyr/samples/coredumps/src/main.c
jmasek: good point, removed

zephyr/samples/coredumps/Kconfig
jmasek: it is actually good idea. will finish it tomorrow

zephyr/samples/coredumps/boards/frdm_rw612.conf
roberthusak, line 1: For documentation purposes, I'd also consider adding a comment to this file mentioning that some options are set in Kconfig.

jmasek: added

zephyr/samples/coredumps/boards/frdm_rw612.conf
```diff
@@ -1,11 +1,10 @@
-CONFIG_WIFI_NXP=y
-CONFIG_NXP_RW610=y
-CONFIG_ETH_DRIVER=n
+# These configs are automatically enabled for FRDM-RW612 board from kconfig
+#WIFI_NXP
+#NXP_WIFI_TX_RX_ZERO_COPY
+#NET_IF_MAX_IPV4_COUNT=y
```
roberthusak:
```suggestion
#NET_IF_MAX_IPV4_COUNT=2
```
?

zephyr/samples/coredumps/prj.conf
```diff
@@ -1,3 +1,11 @@
+# NXP FRDM-RW612 board uses ethernet by default. If you want to use Ethernet instead,
```
roberthusak:
```suggestion
# NXP FRDM-RW612 board uses Wi-Fi by default. If you want to use Ethernet instead,
```
?

zephyr/samples/logs/boards/frdm_rw612.conf
```diff
@@ -1,11 +1,9 @@
-CONFIG_WIFI_NXP=y
-CONFIG_NXP_RW610=y
-CONFIG_ETH_DRIVER=n
+# These configs are automatically enabled for FRDM-RW612 board from kconfig
+#WIFI_NXP
+#NXP_WIFI_TX_RX_ZERO_COPY
+#NET_IF_MAX_IPV4_COUNT=y
```
roberthusak:
```suggestion
#NET_IF_MAX_IPV4_COUNT=2
```
?

zephyr/samples/logs/prj.conf
```diff
@@ -1,3 +1,8 @@
+# NXP FRDM-RW612 board uses ethernet by default. If you want to use Ethernet instead,
```
roberthusak:
```suggestion
# NXP FRDM-RW612 board uses Wi-Fi by default. If you want to use Ethernet instead,
```
?

PR #32, opened 2025-10-02
ESP IDF integration

noorhaq:


noorhaq:
To test it you can open up the folder for the examples or copy it at some other place and build it.
First using `idf.py menuconfig` and in that add the wifi credentials in the example connection configuration and rest of the information related to the spotflow account in the tab under spotflow.  Then build it using the `idf.py build` and then `idf.py flash` to flash it on the board.

PR #33, opened 2025-10-02
Implement setting minimal log severity from cloud

roberthusak:


zephyr/samples/coredumps/src/main.c
roberthusak, line 1: The modifications of this file are just temporary to make the testing of the feature easier. I plan to revert it before merging.

zephyr/src/config/spotflow_config.c
roberthusak, line 1: Move config-related files to `config` folder.

zephyr/src/config/spotflow_config.c
```diff
@@ -0,0 +1,113 @@
...
+	spotflow_config_persistence_try_save(&persisted_settings);
+
+	uint8_t buffer[SPOTFLOW_CONFIG_RESPONSE_MAX_LENGTH];
+	rc = spotflow_config_cbor_encode_reported(&reported_msg, buffer, sizeof(buffer));
+	if (rc < 0) {
+		LOG_ERR("Failed to encode reported configuration response message: %d", rc);
+		return;
+	}
+
+	rc = spotflow_mqtt_publish_config_cbor_msg(buffer, sizeof(buffer));
```
roberthusak: Send via `process_coredumps_or_logs` in `spotflow_processor.c` instead - add a "single-item queue" holding the last version or reported properties to be send.

zephyr/src/config/spotflow_config_persistence.c
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	LOG_INF("Persisted Spotflow configuration loaded");
+}
+
+void spotflow_config_persistence_try_save(struct spotflow_config_persisted_settings* settings)
+{
+	if (settings->contains_sent_log_level) {
+		int rc =
+		    settings_save_one(SPOTFLOW_SETTINGS_PATH_SENT_LOG_LEVEL,
```
roberthusak, line 49: Add comment that checking that the value has changed is handled by Zephyr settings subsystem.

zephyr/src/logging/KConfig
```diff
@@ -54,4 +54,29 @@ config SPOTFLOW_LOGS_PROCESSING_LOG_LEVEL
...
+		Filters the logs sent to Spotflow cloud according to their severity level.
+		This is the default value that can be dynamically overridden from the cloud.
+		Defaults to LOG_MAX_LEVEL, i.e. the maximal (minimal severity) compiled-in level.
+		Levels are:
+		- 0 OFF, do not send
+		- 1 ERROR, send LOG_ERR
+		- 2 WARNING, send LOG_WRN in addition to previous level
+		- 3 INFO, send LOG_INF in addition to previous level
+		- 4 DEBUG, send LOG_DBG in addition to previous level
+
```
noorhaq, line 74: Add LOG_MAX_LEVEL tag information as well. Like it represents the DEBUG or set it to 4.

zephyr/src/logging/spotflow_log_backend.c
```diff
@@ -83,6 +101,10 @@ static void process(const struct log_backend* const backend, union log_msg_gener
 		return;
 	}
 
+	if (log_msg_get_level(log_msg) > spotflow_config_get_sent_log_level()) {
```
noorhaq: I think there might be a logical mistake here, like say it is set to log level info and info message is displayed so won't this cause it to skip?

zephyr/src/net/spotflow_mqtt.c
```diff
@@ -294,11 +350,37 @@ static void mqtt_evt_handler(struct mqtt_client* client, const struct mqtt_evt*
...
+			LOG_ERR("MQTT PUBLISH decode error: %d", evt->result);
+			break;
+		}
+		LOG_DBG("PUBLISH packet id: %u", evt->param.publish.message_id);
+
+		/* The actual topic name is longer to distinguish between different devices */
+		if (mqtt_client_toolset.c2d_message_callback &&
+		    utf8_starts_with(&evt->param.publish.message.topic.topic,
+				     &spotflow_mqtt_config.config_c2d_topic)) {
+			ret = mqtt_read_publish_payload(client, c2d_payload_buffer,
```
noorhaq: I am not sure why are we doing this in publish payload event. Like workflow at this point.

zephyr/src/net/spotflow_processor.c
```diff
@@ -92,11 +92,16 @@ static void process_mqtt()
 	int rc;
 	rc = spotflow_session_metadata_send();
 	if (rc < 0) {
-		LOG_DBG("Failed to send session metadata, aborting MQTT: %d", rc);
+		LOG_WRN("Failed to send session metadata, aborting MQTT: %d", rc);
```
noorhaq: Just a query this WRN will generate another log queue and again the same message will be recevied until the message queue is filled with just this message. So, how are we tackling it here?

zephyr/Kconfig
```diff
@@ -129,4 +129,13 @@ config SPOTFLOW_PROCESSING_THREAD_STACK_SIZE
 		Size of the stack used by the Spotflow processing thread.
 		Increase this value if you experience issues with logs/tls handshake.
 
+config SPOTFLOW_SETTINGS
+	bool "Enable persistence of Spotflow configuration using Zephyr settings subsystem"
+	default y
```
jmasek, line 134: how is it with support on other boards? do they have partition for settings by default? -> is it safe to have it enabled by default?

zephyr/samples/coredumps/prj.conf
```diff
@@ -37,3 +38,10 @@ CONFIG_NET_UDP=y
 CONFIG_WIFI=y
 CONFIG_WIFI_NM=y
 CONFIG_NET_L2_WIFI_MGMT=y
+
+# Persistence of configuration using Zephyr settings
+CONFIG_FLASH=y
+CONFIG_FLASH_MAP=y
+CONFIG_NVS=y
+CONFIG_SETTINGS=y
+CONFIG_SETTINGS_NVS=y
```
jmasek, line 47: shouldn't be our log setting by default enabled only if setting is turned on?

zephyr/src/logging/KConfig
```diff
@@ -54,4 +54,29 @@ config SPOTFLOW_LOGS_PROCESSING_LOG_LEVEL
...
+config SPOTFLOW_LOG_BACKEND_SET_RUNTIME_FILTERING
+	bool "Set runtime filtering when Spotflow log level is set from the cloud"
```
jmasek, line 76: Are there any usages when users do not want to enable this?

zephyr/src/logging/spotflow_log_backend.c
```diff
@@ -69,6 +85,8 @@ static void init(const struct log_backend* const backend)
 	ctx->dropped_backend_count = 0;
 	ctx->message_index = 0;
 
+	spotflow_config_init();
```
jmasek, line 88: it is quite weird that you are calling init -> try load, and conditional compilation is done on the level of loading from disk instead of here. Wouldn't be better to have conditional compilation here and get rid of empty implementations?

jmasek, line 88: it is quite weird that you are calling init -> try load and condifitional compilation is done on the level of loadign from disk instead of here

zephyr/src/config/spotflow_config.c
```diff
@@ -0,0 +1,113 @@
...
+		LOG_ERR("Failed to decode received desired configuration message: %d", rc);
+		return;
+	}
+
+	struct spotflow_config_reported_msg reported_msg = {
+		.contains_acked_desired_config_version = true,
+		.acked_desired_config_version = desired_msg.desired_config_version,
+	};
+
+	struct spotflow_config_persisted_settings persisted_settings = { 0 };
```
jmasek: shouldn't be 0U - setting unsigned int - contains_sent_log_level

zephyr/src/config/spotflow_config_options.c
```diff
@@ -0,0 +1,40 @@
...
+static volatile uint8_t sent_log_level = CONFIG_SPOTFLOW_DEFAULT_SENT_LOG_LEVEL;
+
+uint8_t spotflow_config_get_sent_log_level()
+{
+	return sent_log_level;
+}
+
+void spotflow_config_init_sent_log_level(bool override_value, uint8_t level)
+{
+	if (override_value) {
```
jmasek: does this method makes sense with `override_value` I would prefer to have method `spotflow_config_override_init_sent_log_level` that is called only if override == true

jmasek (changes requested):

zephyr/src/logging/KConfig
roberthusak: Updated, is it better now?

zephyr/Kconfig
roberthusak: `CONFIG_SPOTFLOW_SETTINGS` will be set to its default `y` only if `CONFIG_SETTINGS=y`, because it depends on it. (At least if I understand Kconfig correctly 😅)

zephyr/samples/coredumps/prj.conf
roberthusak: Our `CONFIG_SPOTFLOW_SETTINGS` will be enabled if `CONFIG_SETTINGS` is enabled. Therefore, in the sample, we enable the latter explicitly and we also enable a specific flash settings backend and its dependencies.

zephyr/src/config/spotflow_config.c
roberthusak: `{ 0 }` is here to initiate the struct to an empty default, i.e., all `contains_*` are `false` so nothing is persisted by default. With `spotflow_config_persistence_try_save`, we persist only the options received in the desired config message (only minimal log severity by now, but it might be extended in the future).

But maybe I should rename the variable to `settings_to_persist` or something like that to make the intention clearer.

zephyr/src/logging/KConfig
roberthusak: If they want to manage runtime filtering themselves, i.e., they might want to modify runtime filtering to send DEBUG to our logging backend from MQTT module but have INFO for other modules.

zephyr/src/logging/spotflow_log_backend.c
roberthusak: `spotflow_config_init` does multiple things - it also sets the runtime filter if enabled. I think this way it's encapsulated better.

roberthusak: Duplicate comment.

zephyr/src/config/spotflow_config.c
roberthusak: Regarding `0` vs. `0U` - it shouldn't matter, the first field is `bool` anyway, so both variants should be valid.

zephyr/src/config/spotflow_config_options.c
roberthusak: You're right, `override_value` was weird. I changed the semantics as you suggest and renamed it to something like `init` and `init_default`.

PR #36, opened 2025-10-07
Feature/esp idf v5

noorhaq:
Updated the esp idf component to reflect the changes.


esp_idf/spotflow/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+idf_component_register(SRCS "srcs/net/spotflow_mqtt.c" "srcs/logging/spotflow_log_backend.c" "srcs/logging/spotflow_log_cbor.c" "srcs/logging/spotflow_log_json.c" "spotflow.c"
```
jmasek: naming please use src instead of `srcs`

esp_idf/spotflow/examples/logs/main/app_main.c
```diff
@@ -0,0 +1,64 @@
...
+
+    ESP_ERROR_CHECK(nvs_flash_init());
+    ESP_ERROR_CHECK(esp_netif_init());
+    ESP_ERROR_CHECK(esp_event_loop_create_default());
+
+    /* This helper function configures Wi-Fi or Ethernet, as selected in menuconfig.
+     * Read "Establishing Wi-Fi or Ethernet Connection" section in
+     * examples/protocols/README.md for more information about this function.
+     */
+    ESP_ERROR_CHECK(example_connect());
```
jmasek, line 36: is this implemented? Where does the `example_connect` come from?

esp_idf/spotflow/examples/logs/sdkconfig.ci
```diff
@@ -0,0 +1,20 @@
+CONFIG_MQTT_USE_CUSTOM_CONFIG=y
+CONFIG_MQTT_TCP_DEFAULT_PORT=1883
+CONFIG_MQTT_SSL_DEFAULT_PORT=8883
+CONFIG_MQTT_WS_DEFAULT_PORT=80
+CONFIG_MQTT_WSS_DEFAULT_PORT=443
```
jmasek, line 5: what are these configs? when they are used?

esp_idf/spotflow/src/logging/spotflow_log_backend.c
```diff
@@ -0,0 +1,60 @@
...
+int spotflow_log_backend(const char *fmt, va_list args)
+{
+    // Format the log string into a buffer
+    char buffer[CONFIG_SPOTFLOW_LOG_BUFFER_SIZE];
+    int len = vsnprintf(buffer, sizeof(buffer), fmt, args);
+
+#if CONFIG_USE_JSON_PAYLOAD
+    char *log_severity = NULL;
+    if (len > 0 && len < CONFIG_SPOTFLOW_LOG_BUFFER_SIZE) {
+        switch (buffer[0]) {
```
jmasek: why it is here and not in the log_json or log_cbor file?

esp_idf/spotflow/examples/logs/main/app_main.c
```diff
@@ -0,0 +1,64 @@
...
+
+    spotflow_init();
+
+    int test_num = 25;
```
roberthusak: Not used.

esp_idf/spotflow/spotflow.c
```diff
@@ -0,0 +1,19 @@
...
+#endif
+
+
+vprintf_like_t original_vprintf = NULL;
+
+// Initilize the spotlow log backend
+/**
+ * @brief 
+ * @details To utilize the esp_log_set_vprintf function to expose the logs
+ */
```
roberthusak: Unify comment types.

esp_idf/spotflow/src/logging/spotflow_log_backend.c
```diff
@@ -0,0 +1,60 @@
...
+    }
+#else
+    uint8_t log_severity = 0;
+    if (len > 0 && len < CONFIG_SPOTFLOW_LOG_BUFFER_SIZE) {
+        switch (buffer[0]) {
+            case 'I': log_severity = 0x28;  break; //Info
+            case 'D': log_severity = 0x1E; break; //Debug
+            case 'W': log_severity = 0x32; break; //Warning
+            case 'E': log_severity = 0x3C; break; //Error
+            case 'V': log_severity = 0x28; break; //Verbose right now set to info
```
roberthusak: Reorder with either increasing or decreasing severity. Verbose should be mapped to Debug (Debug is more detailed than Info).

esp_idf/spotflow/src/logging/spotflow_log_backend.c
```diff
@@ -0,0 +1,60 @@
...
+
+#if CONFIG_USE_JSON_PAYLOAD
+    char *log_severity = NULL;
+    if (len > 0 && len < CONFIG_SPOTFLOW_LOG_BUFFER_SIZE) {
+        switch (buffer[0]) {
+            case 'I': log_severity = "INFO"; break;
+            case 'D': log_severity = "DEBUG"; break;
+            case 'W': log_severity = "WARNING"; break;
+            case 'E': log_severity = "ERROR"; break;
+            case 'V': log_severity = "VERBOSE"; break;
```
roberthusak: Map Verbose to Debug.

esp_idf/spotflow/src/logging/spotflow_log_cbor.c
```diff
@@ -0,0 +1,235 @@
...
+        body[body_len - 1] = '\0';
+    }
+
+    
+    uint32_t uptime_ms = esp_log_timestamp();
+
+    // Get device timestamp (UNIX epoch in ms)
+    struct timespec ts;
+    clock_gettime(CLOCK_REALTIME, &ts);
+    uint64_t timestamp_ms = (uint64_t)ts.tv_sec * 1000 + ts.tv_nsec / 1000000;
```
roberthusak: Unused, CBOR message should contain it (and uptime as well).

Also, is there any way to check whether `clock_gettime(CLOCK_REALTIME, &ts)` returns proper value (e.g., from NTP), before we send it?

esp_idf/spotflow/src/logging/spotflow_log_cbor.c
```diff
@@ -0,0 +1,235 @@
...
+    // Buffer to create array to cointain several items
+    CborEncoder array_encoder;
+    CborEncoder map_encoder;
+    uint8_t *buf = malloc(CONFIG_SPOTFLOW_CBOR_LOG_MAX_LEN);
+    if (!buf) {
+        SPOTFLOW_LOG("Failed to allocate CBOR buffer");
+        return NULL;
+    }
+    
+    char *separator = strchr(body, ':');
```
roberthusak: What's before the colon?

esp_idf/spotflow/srcs/net/spotflow_mqtt.c
```diff
@@ -0,0 +1,94 @@
...
+
+esp_mqtt_client_handle_t client = NULL;
+bool mqtt_connected = false;
+
+#if CONFIG_BROKER_CERTIFICATE_OVERRIDDEN == 1
+static const uint8_t mqtt_spotflow_io_pem_start[]  = "-----BEGIN CERTIFICATE-----\n" CONFIG_BROKER_CERTIFICATE_OVERRIDE "\n-----END CERTIFICATE-----";
+#else
+extern const uint8_t mqtt_spotflow_io_pem_start[]   asm("_binary_mqtt_spotflow_io_pem_start");
+#endif
+extern const uint8_t mqtt_spotflow_io_pem_end[]   asm("_binary_mqtt_spotflow_io_pem_end");
```
roberthusak: `mqtt_spotflow_io_pem_end` is unused (and incorrect if `CONFIG_BROKER_CERTIFICATE_OVERRIDDEN == 1`).

esp_idf/spotflow/srcs/net/spotflow_mqtt.c
```diff
@@ -0,0 +1,94 @@
...
+{
+    // ESP_LOGD(TAG, "Event dispatched from event loop base=%s, event_id=%" PRIi32, base, event_id);
+    esp_mqtt_event_handle_t event = event_data;
+    // esp_mqtt_client_handle_t client = event->client;
+    int msg_id;
+    switch ((esp_mqtt_event_id_t)event_id) {
+    case MQTT_EVENT_CONNECTED:
+        SPOTFLOW_LOG("MQTT_EVENT_CONNECTED");
+        mqtt_connected = true;
+        esp_mqtt_client_subscribe(client, "ingest-json", 1);
```
roberthusak: Unnecessary to subscribe to this topic - we're just sending messages there (and only in the JSON case), not consuming them.

esp_idf/spotflow/src/logging/spotflow_log_backend.c
```diff
@@ -0,0 +1,60 @@
+#include "logging/spotflow_log_backend.h"
+
+static const char *TAG = "spotflow_testing";
+#include <stdarg.h>
+
+int spotflow_log_backend(const char *fmt, va_list args)
+{
+    // Format the log string into a buffer
+    char buffer[CONFIG_SPOTFLOW_LOG_BUFFER_SIZE];
```
roberthusak: Is it safe to allocate 512 bytes (or more if configured) on the stack and not on the heap?

esp_idf/spotflow/CMakeLists.txt
roberthusak, line 1: Is it necessary to have `spotflow` in the path? I.e., would it be possible to have everything just under `esp_idf` as in the Zephyr case?

esp_idf/spotflow/include/logging/spotflow_log_backend.h
```diff
@@ -0,0 +1,13 @@
+#ifndef SPOTFLOW_LOG_H
+#define SPOTFLOW_LOG_H
+
+#include "spotflow.h"
+
+int spotflow_log_backend(const char *fmt, va_list args);
+#if CONFIG_SPOTFLOW_DEBUG_MESSAGE_TERMINAL
+    #define SPOTFLOW_LOG(fmt, ...) printf("[SPOTFLOW] " fmt, ##__VA_ARGS__)
```
roberthusak, line 14: Would it be possible to sent the messages from our module to the cloud as well? (Instead of just printing them to the console.)

esp_idf/spotflow/mqtt_spotflow_io.pem
```diff
@@ -0,0 +1,31 @@
+-----BEGIN CERTIFICATE-----
```
jmasek, line 1: please add comment - or change the name so it is obvious that we are using lets encrypt x1 root certificate

esp_idf/spotflow/examples/logs/dependencies.lock
```diff
@@ -0,0 +1,39 @@
...
+    version: '*'
+  spotflow:
+    component_hash: 830fef9f79db5f0f4beb15084565b8fef376afd9b632704a44674f9a09dec45f
+    dependencies:
+    - name: idf
+      version: '>=5.0'
+    - name: espressif/cbor
+      version: ^0.6.1~3
+    source:
+      git: https://github.com/noorhaq/device-sdk.git
```
roberthusak: We mustn't forget to turn these links into https://github.com/spotflow-io/device-sdk.git eventually.

esp_idf/spotflow/examples/logs/sdkconfig.ci
```diff
@@ -0,0 +1,20 @@
...
+CONFIG_MBEDTLS_SSL_OUT_CONTENT_LEN=4096
+CONFIG_EXAMPLE_CONNECT_ETHERNET=y
+CONFIG_EXAMPLE_CONNECT_WIFI=n
+CONFIG_EXAMPLE_USE_INTERNAL_ETHERNET=y
+CONFIG_EXAMPLE_ETH_PHY_IP101=y
+CONFIG_EXAMPLE_ETH_MDC_GPIO=23
+CONFIG_EXAMPLE_ETH_MDIO_GPIO=18
+CONFIG_EXAMPLE_ETH_PHY_RST_GPIO=5
+CONFIG_EXAMPLE_ETH_PHY_ADDR=1
+CONFIG_EXAMPLE_CONNECT_IPV6=y
```
roberthusak, line 19: How are these `CONFIG_EXAMPLE_*` options used?

esp_idf/spotflow/src/net/spotflow_mqtt.c
```diff
@@ -0,0 +1,94 @@
+
+#include "net/spotflow_mqtt.h"
+
+static const char *TAG = "spotflow_mqtt";
+
+esp_mqtt_client_handle_t client = NULL;
+bool mqtt_connected = false;
```
jmasek: Shouldn't be `mqtt_connected` atomic?

esp_idf/spotflow/src/logging/spotflow_log_cbor.c
```diff
@@ -0,0 +1,235 @@
...
+{
+    while (len--) {
+        SPOTFLOW_LOG("%02X ", *buf++);
+    }
+}
+
+/**
+ * Decode CBOR data manually
+ */
+static CborError example_dump_cbor_buffer(CborValue *it, int nestingLevel)
```
roberthusak: Is it necessary to have this function in the code? Wouldn't it be cleaner to just output the hex representation to the console and debug the CBOR format only if necessary and with tools outside the source code?

roberthusak (commented):
Hi @noorhaq, thank you for the PR! I'm the other maintainer of the device SDK and I added some comments. @jmasek will also add issues he found.

esp_idf/spotflow/include/net/spotflow_mqtt.h
```diff
@@ -0,0 +1,11 @@
+#ifndef SPOTFLOW_MQTT_H
+#define SPOTFLOW_MQTT_H
+
+#include "spotflow.h"
+
+extern esp_mqtt_client_handle_t client;
+extern bool mqtt_connected;
+
+void mqtt_app_start(void);
```
jmasek, line 10: Who calls this? the `example_connect` implementation? is it some esp-idf convention?

esp_idf/spotflow/src/logging/spotflow_log_backend.c
```diff
@@ -0,0 +1,60 @@
...
+        switch (buffer[0]) {
+            case 'I': log_severity = "INFO"; break;
+            case 'D': log_severity = "DEBUG"; break;
+            case 'W': log_severity = "WARNING"; break;
+            case 'E': log_severity = "ERROR"; break;
+            case 'V': log_severity = "VERBOSE"; break;
+            default: log_severity = "NONE"; break;
+        }
+
+        if(mqtt_connected)
```
jmasek: This means that if the connection is dropped, the message will get lost, is it correct? In our Zephyr SDK, ring buffer is used so the messages can be buffered until the connection is restored. 
Also, I see that the messages sent before setting up our backend are lost. It can be problematic since the developers can miss important information from the MCU start.

jmasek (changes requested):

esp_idf/spotflow/src/logging/spotflow_log_cbor.c
noorhaq: Yes like I think it is in UTC 0 but for now we are not using it. As this will be a on a separate PR like the metadata being sent. Right now these were just added to test and afterwards map them to correct ones like uptime, actual time etc.

esp_idf/spotflow/src/logging/spotflow_log_backend.c
noorhaq: These are default ones like the one ESP_Log library offers like if you use the ESP_LOGV it will be on verbose same for ESP_LOGD and this is also controlled through the logging level of the esp internal library if we change it here it will cause confusion as which log type is declared and which is sent.

esp_idf/spotflow/src/logging/spotflow_log_backend.c
noorhaq: I will change this.

esp_idf/spotflow/src/logging/spotflow_log_cbor.c
noorhaq: Some data like time and stuff which is currently not supported by the cloud so I am stripping them to make it more in line with the zephyr logs so cloud recognizes them.

esp_idf/spotflow/srcs/net/spotflow_mqtt.c
noorhaq: Yep that kind of left during debug I will remove it. BTW don't we need to subscribe to it in case to increase log severity or any case?

esp_idf/spotflow/CMakeLists.txt
noorhaq: Yes it is as it will be downloaded as a component named as `spotflow`.
Like when you will test you will see a folder being created called managed_components by `esp idf` and in that it will pull all the dependencies components `spotflow` being one

esp_idf/spotflow/include/logging/spotflow_log_backend.h
noorhaq: Yes it is right now this was also a test case because in zephyr we are just keeping the log level higher so it was kind of pilot to make sure we can control these logs without impacting anything else. Now only issue is how to resolve recursion i.e. we are sending a message -> debug log of that message (Sending message) creates another event to send message and so on. Until we are out of memory and only logs are just Sending message.

esp_idf/spotflow/examples/logs/dependencies.lock
noorhaq: Yes I had mentioned this to Jan as well like when our PR is merged the next change is just to use the commit hash and change this link here. Right now it is just a placeholder for test

esp_idf/spotflow/examples/logs/sdkconfig.ci
noorhaq: ESP IDF has a example component which handles wifi connections. So in example instead of creating our own implementation of this logic I just used that one as customers will already have their own wifi configuration so this is only limited to example. Not in our components.

esp_idf/spotflow/src/logging/spotflow_log_cbor.c
noorhaq: We can remove it. I just kept it as placeholder in case we need to debug as it makes it easier.

esp_idf/spotflow/examples/logs/main/app_main.c
noorhaq: It comes from esp idf own implementaion in idf_component.yml you can see it in the dependencies.

esp_idf/spotflow/src/logging/spotflow_log_backend.c
noorhaq: Because it is forming the message before sending it to either one of them. Like striping the message and making it conform to our pattern. I did the both here for simplicity.

esp_idf/spotflow/src/net/spotflow_mqtt.c
noorhaq: Only one thread was accessing it and only one will so I didn't feel the need for it to be atomic.

esp_idf/spotflow/src/logging/spotflow_log_backend.c
noorhaq: Yes that's like in the second PR as this was just a implementation to make the base level work. that's why it's missing the queue and metadata transmission.

esp_idf/spotflow/examples/logs/main/app_main.c
noorhaq: Fixed.

noorhaq: The test_num is used inside the example like the while(1)

roberthusak: Yeah, sorry, my comment was wrong. It's used indeed. It just seems illogical to me why it starts at 25. Shouldn't it start at 1?

esp_idf/spotflow/include/logging/spotflow_log_backend.h
roberthusak: OK 👍 Yeah, the recursion problem makes sense, log levels will most likely play a role in preventing it. We can discuss it tomorrow.

esp_idf/spotflow/src/logging/spotflow_log_backend.c
roberthusak: This comment was meant to ensure consistency with the encoding to CBOR below. Our backend doesn't recognize the level `"VERBOSE"` (at least as of now) and `"DEBUG"` is the closest one to it. We can discuss it tomorrow.

esp_idf/spotflow/srcs/net/spotflow_mqtt.c
roberthusak: We've got a different topic for that: `config-cbor-c2d`, more details later.

esp_idf/spotflow/src/logging/spotflow_log_cbor.c
jmasek: isn't there source - module or c file? because we normally push to the cloud in labels as source field

esp_idf/spotflow/examples/logs/sdkconfig.ci
jmasek: the example_connect I expect, right? :)

esp_idf/spotflow/include/net/spotflow_mqtt.h
noorhaq: Yes.

esp_idf/spotflow/mqtt_spotflow_io.pem
noorhaq: Changed the name.

esp_idf/spotflow/src/logging/spotflow_log_backend.c
noorhaq: Moved to their respective files.

esp_idf/spotflow/CMakeLists.txt
noorhaq: Done.

noorhaq:
Hi,
 I have addressed and fixed I think mostly all the issues. Plus added the metadata too. Their is some issue in their like I have decoded the CBOR data and sequence number is being sent but on the cloud side it is not if I comment out the labels field and reduce the CBOR count to 6 instead of 7 with the field it receives the sequence number. Not sure what might be going wrong here.

esp_idf/spotflow/examples/logs/main/app_main.c
```diff
@@ -0,0 +1,64 @@
...
+#include "esp_partition.h"
+#include "nvs_flash.h"
+#include "esp_event.h"
+#include "esp_netif.h"
+#include "protocol_examples_common.h"
+
+#include "esp_log.h"
+#include "mqtt_client.h"
+#include "esp_tls.h"
+#include "esp_ota_ops.h"
```
jmasek: unused imports

noorhaq:
Added Queue for message retention as well.

esp_idf/spotflow/Kconfig.projbuild
```diff
@@ -0,0 +1,77 @@
...
+    config SPOTFLOW_SERVER_HOSTNAME
+        string "Hostname of Spotflow observability platform"
+        default "mqtts://mqtt.spotflow.io:8883"
+        help
+            URL of an mqtt broker which spotflow connects to.
+    
+    config SPOTFLOW_DEVICE_ID
+        string "Device ID device will use to connect to Spotflow observability platform"
+        help
+            Device ID is a unique identifier for your device in Spotflow observability
```
jmasek, line 12: is automatic device id resolution implemented?

esp_idf/spotflow/Kconfig.projbuild
```diff
@@ -0,0 +1,77 @@
...
+        default y if BROKER_CERTIFICATE_OVERRIDE != ""
+
+    
+    config USE_JSON_PAYLOAD
+        bool "Use json payload instead of cbor"
+        default n
+        help
+            By default cbor format is used. If json format is required this can be set to use instead of cbor.
+    
+    config SPOTFLOW_GENERATE_BUILD_ID
```
jmasek, line 32: not implemented yet, correct? Maybe could you add note that results in no-op?

esp_idf/spotflow/Kconfig.projbuild
```diff
@@ -0,0 +1,77 @@
...
+	bool "Generate build ID"
+	default n
+	# Use binary descriptors when possible
+	help
+		Generate build ID and embed it as a binary descriptor.
+		Build ID allows to match core dumps with the original .elf file.
+
+    config SPOTFLOW_LOG_BACKEND
```
jmasek, line 40: not implemented yet, correct? Maybe could you add note that results in no-op?

esp_idf/spotflow/include/logging/spotflow_log_cbor.h
```diff
@@ -0,0 +1,10 @@
+#ifndef SPOTFLOW_LOG_CBOR_H
+#define SPOTFLOW_LOG_CBOR_H
+
+#include "spotflow.h"
+#include "logging/spotflow_log_backend.h"
+
+uint8_t* log_cbor(char *fem, char* body, uint8_t severity, size_t *out_len, struct message_metadata *metadata);
```
jmasek: can be static?

esp_idf/spotflow/include/logging/spotflow_log_json.h
```diff
@@ -0,0 +1,8 @@
+#ifndef SPOTFLOW_LOG_JSON_H
+#define SPOTFLOW_LOG_JSON_H
+
+#include "spotflow.h"
+
+char* log_json(char* body, const char* severity);
```
jmasek: can be static?

esp_idf/spotflow/include/logging/spotflow_log_json.h
```diff
@@ -0,0 +1,8 @@
...
+void log_json_send(char* buffer,char log_sevirity);
```
jmasek: typo severity,

esp_idf/spotflow/src/logging/spotflow_log_json.c
```diff
@@ -0,0 +1,79 @@
...
+    );
+    return json_str;
+}
+
+/**
+ * @brief Form and send the JSON parameters
+ * 
+ * @param buffer 
+ */
+void log_json_send(char* buffer,char log_sevirity)
```
jmasek: log_sevirity parameter not used

jmasek: it is quite strange behavior that if you enable the json you will lose the buffering functionality

esp_idf/spotflow/src/logging/spotflow_log_cbor.c
```diff
@@ -0,0 +1,147 @@
...
+
+        queue_push((const char*) clog_cbor, len);
+
+        if(atomic_load(&mqtt_connected))
+        {
+            char *queue_buffer = malloc(CONFIG_SPOTFLOW_CBOR_LOG_MAX_LEN);
+
+            while (queue_read(queue_buffer) != -1)
+            {
+                esp_mqtt_client_publish(client, "ingest-cbor", (const char*)queue_buffer , CONFIG_SPOTFLOW_CBOR_LOG_MAX_LEN, 1, 0); // Treat it as a NULL terminated string
```
jmasek: what happens if during sending client gets disconnected? will it be handled gracefully?
No response code from publish is processed
Is ok that the caller of logging function (user) will be blocked until all messages are published? I expect publish is some async function so the delay may not be be significant...

esp_idf/spotflow/src/logging/spotflow_log_queue.c
```diff
@@ -0,0 +1,86 @@
+
+#include "logging/spotflow_log_queue.h"
+
+mqd_t write_descr;
```
jmasek: did you consider using `xQueue` there would be less allocations/frees

esp_idf/spotflow/src/logging/spotflow_log_queue.c
```diff
@@ -0,0 +1,86 @@
...
+ */
+void queue_push(const char *msg, size_t len)
+{
+    if (mq_send(write_descr, msg, len, 0) < 0) {
+        if (errno == EAGAIN) {
+            // queue full — remove oldest message
+            char dropped[CONFIG_SPOTFLOW_CBOR_LOG_MAX_LEN];
+            ssize_t len = mq_receive(write_descr, dropped, sizeof(dropped), NULL);
+            if (len >= 0) {
+                dropped[len] = '\0';
```
jmasek: why you are doing this? can you please add comment to code?

esp_idf/spotflow/src/logging/spotflow_log_backend.c
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    // Create a copy of va_list args
+    va_list args_copy;
+    va_copy(args_copy, args);
+
+    int len = vsnprintf(NULL, 0, fmt_copy, args_copy);  // Get the required length
+
+    char log_severity = fmt_copy[0];
+    for(const char *p = fmt_copy; *p || *p == ':'; p++)
```
jmasek, line 33: please add comment explaining what you are really doing
like - logs are provided by esp-idf in the following format -> extracting log level, seq number, etc

esp_idf/spotflow/include/logging/spotflow_log_cbor.h
jmasek: Most of the parameters can be const, aren't they?

jmasek (changes requested):

esp_idf/spotflow/src/logging/spotflow_log_queue.c
noorhaq: Yes but this is like has a Queue Message priority index i.e. we can send some high priority logs or save them while dropping low priority ones.

noorhaq: So that's why I used this one instead of xQueue as this will provide us with Message Priority parameter which we are not using right now but can in future implementations. To make sure high priority errors don't get dropped due to low bandwidth or weak signal areas

esp_idf/spotflow/Kconfig.projbuild
noorhaq: I will add it.

esp_idf/spotflow/Kconfig.projbuild
noorhaq: Can you share the format like Prefix_Mac_Address.

esp_idf/spotflow/examples/logs/main/app_main.c
noorhaq: Done.

esp_idf/spotflow/Kconfig.projbuild
noorhaq: Done.

esp_idf/spotflow/Kconfig.projbuild
noorhaq: Right now the Device ID is only MAC Address like if not set in menuconfig.

esp_idf/spotflow/src/logging/spotflow_log_cbor.c
noorhaq: Well we don't have to if say the client publish is effected the event is generated for the MQTT disconnect and the event disconnect stops it. Like those events are listened and processed separately.  I will add a condition here too so that the async behaviour of MQTT disconnect won't cause any issues here.

esp_idf/spotflow/src/logging/spotflow_log_json.c
noorhaq: Yes Json was kind of a concept implementation outside the scope of current work that's why my main focus was to have the current setup processed before adding this. Because in Json we don't have Coredumps functionality too. Like for that we will be using CBOR.

esp_idf/spotflow/include/logging/spotflow_log_cbor.h
noorhaq: Done.

esp_idf/spotflow/src/logging/spotflow_log_cbor.c
noorhaq: Added a Check in the loop condition as well.

esp_idf/spotflow/src/logging/spotflow_log_queue.c
noorhaq: Added.

esp_idf/spotflow/src/logging/spotflow_log_queue.c
jmasek: ok, lets keep it that way and we can make the decision based on the product roadmap in the future.

esp_idf/spotflow/src/logging/spotflow_log_json.c
jmasek: discused in chat

esp_idf/spotflow/include/logging/spotflow_log_cbor.h
```diff
@@ -0,0 +1,10 @@
+#ifndef SPOTFLOW_LOG_CBOR_H
+#define SPOTFLOW_LOG_CBOR_H
+
+#include "spotflow.h"
+#include "logging/spotflow_log_backend.h"
+
+uint8_t* log_cbor(const char *fem, char* body, const uint8_t severity, size_t *out_len, const struct message_metadata *metadata);
```
jmasek: what is fem, can you please choose more descriptive name?

esp_idf/spotflow/include/logging/spotflow_log_queue.h
```diff
@@ -0,0 +1,23 @@
+#ifndef SPOTFLOW_LOG_QUEUE_H
+#define SPOTFLOW_LOG_QUEUE_H
+
+#include <stdio.h>
+#include <fcntl.h>
+#include <sys/stat.h>
+#include <unistd.h>
+#include <stdbool.h>
+#include <pthread.h>
+#include <mqueue.h>
```
jmasek: Why are these imports there when they are used in c file? in header files, there should generally be only things that are needed for declaration.
Generally, dependencies in header files should be minimized.

jmasek (changes requested):

esp_idf/spotflow/include/spotflow.h
```diff
@@ -0,0 +1,29 @@
...
+#include "esp_mac.h"
+#include "mqtt_client.h"
+#include "esp_tls.h"
+#include <sys/param.h>
+#include <time.h>
+#include "esp_log_timestamp.h"
+
+#include "net/spotflow_mqtt.h"
+#include "logging/spotflow_log_backend.h"
+#include "logging/spotflow_log_queue.h"
```
jmasek: unused imports

jmasek (changes requested):

esp_idf/spotflow/src/logging/spotflow_log_cbor.c
```diff
@@ -0,0 +1,154 @@
...
+        switch (log_severity) {
+            case 'E': severity = 0x3C; break; //Error
+            case 'W': severity = 0x32; break; //Warning
+            case 'I': severity = 0x28;  break; //Info
+            case 'D': severity = 0x1E; break; //Debug
+            case 'V': severity = 0x1E; break; //Verbose right now set to debug
+            default: severity = 0x0; break; //In case no log type set it to 0, unknown level
+        }
+
+        size_t len;
```
jmasek: shadowed variable - please rename, validate rest of the code uses the correct one.

esp_idf/spotflow/src/logging/spotflow_log_cbor.c
```diff
@@ -0,0 +1,154 @@
...
+
+        queue_push((const char*) clog_cbor, len);
+
+        if(atomic_load(&mqtt_connected))
+        {
+            char *queue_buffer = malloc(CONFIG_SPOTFLOW_CBOR_LOG_MAX_LEN);
+
+            while (queue_read(queue_buffer) != -1 && atomic_load(&mqtt_connected)) //Check if mqtt disconnect event is not generated.
+            {
+                esp_mqtt_client_publish(client, "ingest-cbor", (const char*)queue_buffer , len, 1, 0); // Sending MQTT message the CONFIG_SPOTFLOW_CBOR_LOG_MAX_LEN provides the buffer length
```
jmasek: not checking response code

esp_idf/spotflow/src/logging/spotflow_log_cbor.c
```diff
@@ -0,0 +1,154 @@
...
+        size_t len;
+        uint8_t *clog_cbor = log_cbor(fmt, buffer, severity, &len, metadata);
+
+        queue_push((const char*) clog_cbor, len);
+
+        if(atomic_load(&mqtt_connected))
+        {
+            char *queue_buffer = malloc(CONFIG_SPOTFLOW_CBOR_LOG_MAX_LEN);
+
+            while (queue_read(queue_buffer) != -1 && atomic_load(&mqtt_connected)) //Check if mqtt disconnect event is not generated.
```
jmasek: you are not sending actual buffer length but length of the serialized message.

esp_idf/spotflow/src/logging/spotflow_log_cbor.c
```diff
@@ -0,0 +1,154 @@
...
+ * 
+ * @param buffer 
+ */
+void log_cbor_send(const char *fmt, char* buffer, const char log_severity, const struct message_metadata *metadata)
+{
+    int len = strlen(buffer);
+    uint8_t severity = 0;
+    if (len > 0 && len < CONFIG_SPOTFLOW_LOG_BUFFER_SIZE) {
+        switch (log_severity) {
+            case 'E': severity = 0x3C; break; //Error
```
jmasek: please use enum for better readability
```
typedef enum {
    LOG_SEVERITY_ERROR = 0x3C,
    LOG_SEVERITY_WARN  = 0x32,
    LOG_SEVERITY_INFO  = 0x28,
    LOG_SEVERITY_DEBUG = 0x1E,
} LogSeverity;

```

esp_idf/spotflow/src/logging/spotflow_log_backend.c
```diff
@@ -0,0 +1,131 @@
...
+
+    char *fmt_copy_1 = fmt_copy; // Just a location for the original memory to free later.
+    char *separator = strchr(fmt_copy, ':');
+    if (separator != NULL) {
+        // Start the body from the character after the colon
+        fmt_copy = separator + 2;
+    }
+
+    // If the len is smaller than 0 or if the log is bigger than the set buffer size.
+    if (len < 0 || len > CONFIG_SPOTFLOW_LOG_BUFFER_SIZE) {
```
jmasek: should be if (len < 0 || len >= CONFIG_SPOTFLOW_LOG_BUFFER_SIZE) because you use
vsnprintf(buffer, len+1

jmasek (changes requested):

esp_idf/spotflow/src/logging/spotflow_log_cbor.c
noorhaq: Done resolved.

esp_idf/spotflow/src/logging/spotflow_log_cbor.c
noorhaq: Added.

esp_idf/spotflow/Kconfig.projbuild
```diff
@@ -0,0 +1,70 @@
...
+        help
+            URL of an mqtt broker which spotflow connects to.
+    
+    config SPOTFLOW_DEVICE_ID
+        string "Device ID device will use to connect to Spotflow observability platform"
+        help
+            Device ID is a unique identifier for your device in Spotflow observability
+            platform. If empty, the hexadecimal representation of the result of
+            hwinfo_get_device_id(...) will be used. You can provide the device ID
+            in the code by defining the function char* spotflow_override_device_id(). Keep it less than 32 characters.
```
roberthusak: I thinks the description should be updated: `hwinfo_get_device_id` is specific to Zephyr and the weakly defined `spotflow_override_device_id` isn't currently implemented for ESP-IDF, right?

esp_idf/spotflow/Kconfig.projbuild
roberthusak: Please also mention in the help of the Kconfig option below.

esp_idf/spotflow/Kconfig.projbuild
roberthusak: Please also mention in the help of the Kconfig option below.

esp_idf/spotflow/src/logging/spotflow_log_queue.c
```diff
@@ -0,0 +1,98 @@
...
+ * @brief To Add a message in Queue
+ * 
+ * @param msg Log Message 
+ */
+void queue_push(const char *msg, size_t len)
+{
+    if (mq_send(write_descr, msg, len, 0) < 0) { // Check if their is no error while writing to the queue buffer.
+        if (errno == EAGAIN) {                  // In case the Queue Buffer is full remove the oldest message
+            // queue full — remove oldest message
+            char dropped[CONFIG_SPOTFLOW_CBOR_LOG_MAX_LEN]; //Create a temporary array of msg size.
```
roberthusak: Is it safe to reserve this amount of memory (1kiB by default) on the stack? And is it even necessary? Can't we just drop the message without copying its contents?

esp_idf/spotflow/src/logging/spotflow_log_queue.c
```diff
@@ -0,0 +1,98 @@
...
+}
+
+/**
+ * @brief To read the queue message
+ * 
+ * @param buffer 
+ * @return int In case of error it returns -1 otherwise it returns the actual buffer_size
+ */
+
+int queue_read(char *buffer)
```
roberthusak: Wouldn't it be more conventional (and also easier to check that the usage is correct) to add the size of the buffer as the second parameter?

esp_idf/spotflow/src/net/spotflow_mqtt.c
```diff
@@ -0,0 +1,109 @@
...
+void mqtt_app_start(void)
+{
+
+    static char device_id[32]; // For saving the device ID
+    uint8_t mac[6]; //The Mac Address string
+    esp_read_mac(mac, ESP_MAC_WIFI_STA); // Read the mac Address
+    snprintf(device_id, sizeof(device_id), "%s",
+            strlen(CONFIG_SPOTFLOW_DEVICE_ID) ? CONFIG_SPOTFLOW_DEVICE_ID :
+            ({ static char tmp[32]; snprintf(tmp, sizeof(tmp), "%02X%02X%02X%02X%02X%02X",
+                                                mac[0], mac[1], mac[2], mac[3], mac[4], mac[5]); tmp; }));  // Check if CONFIG_SPOTFLOW_DEVICE_ID is empty if no then copy the device ID, if yes then generate it.
```
roberthusak: I'm quite confused by this long expression calling `snprintf`, especially by those two last lines. Can you restructure the code to make it more readable?

esp_idf/spotflow/Kconfig.projbuild
noorhaq: Done.

esp_idf/spotflow/src/logging/spotflow_log_queue.c
noorhaq: Sadly No. This is the most efficient way. Most we can do is move the memory from stack to heap.

esp_idf/spotflow/src/logging/spotflow_log_queue.c
noorhaq: I made it so it's easier for the receiving end to verify as we will need to return something like a error code or the actual buffer size. Like I copyed the mq_receive format here. If successful it returns the bytes used otherwise it returns a error code.

esp_idf/spotflow/src/net/spotflow_mqtt.c
noorhaq: I tried to make it one liner. What it does is using the low format if else function.
It is simple like 
strlen(CONFIG_SPOTFLOW_DEVICE_ID) // This is the expression which we need to check. If empty i.e. 0 or false
 ? CONFIG_SPOTFLOW_DEVICE_ID // This will only happen if true i.e. copy the contents inside the device ID
:
            ({ static char tmp[32]; snprintf(tmp, sizeof(tmp), "%02X%02X%02X%02X%02X%02X",
                                                mac[0], mac[1], mac[2], mac[3], mac[4], mac[5]); tmp; })) //else check the device MAC and set it.

esp_idf/spotflow/examples/logs/main/idf_component.yml
```diff
@@ -0,0 +1,7 @@
+dependencies:
+  protocol_examples_common:
+    path: ${IDF_PATH}/examples/common_components/protocol_examples_common
+  spotflow:
+    # git: https://github.com/noorhaq/device-sdk.git
+    path: /home/noor/spotflow/device-sdk/esp_idf/spotflow
```
jmasek: it seems your personal setting leaked to the PR

esp_idf/spotflow/src/logging/spotflow_log_cbor.c
```diff
@@ -0,0 +1,148 @@
...
+        switch (log_severity) {
+            case 'E': severity = LOG_SEVERITY_ERROR; break; //Error
+            case 'W': severity = LOG_SEVERITY_WARN; break; //Warning
+            case 'I': severity = LOG_SEVERITY_INFO;  break; //Info
+            case 'D': severity = LOG_SEVERITY_DEBUG; break; //Debug
+            case 'V': severity = LOG_SEVERITY_DEBUG; break; //Verbose right now set to debug
+            default: severity = 0x0; break; //In case no log type set it to 0, unknown level
+        }
+
+        uint8_t *clog_cbor = log_cbor(fmt, buffer, severity, &len, metadata); // It reuses the length
```
jmasek, line 142: Why are you reusing the variable - it is generally considered a bad practice - Misra and cert c discourage it in various places.

esp_idf/spotflow/src/logging/spotflow_log_queue.c
roberthusak: I see, that's a pitty 🫤 I'm afraid that it might lead to stack overflows in edge cases (when the user increases `CONFIG_SPOTFLOW_CBOR_LOG_MAX_LEN` too much and the dropping is suddenly needed). And I'd rather avoid dynamic allocation because of the possibility that such a memory block is not available (e.g., due to fragmentation).

Is it possible to specify the size of the stack of the thread calling this function? (so that we could add `CONFIG_SPOTFLOW_CBOR_LOG_MAX_LEN` to it) Or maybe we can make the variable static - that would potentially consume more memory, but it would be more deterministic.

@jmasek: What's your take on this?

jmasek: I would prefer use a pointer-based queue (passing by reference). Noor is using posix based queue that is priority queue that could be helpful in the future. The disadvantage is that mqd_t always copies the data - it is made mainly for message passing rather than really passing the data. 
It would be possible to use default FreeRTOS queues as xQueue but it does not have priorities and we do not use them right now Also in zephyr we use queue without support for priorities.

noorhaq:
I am trying another model like instead of saving the messages in the queue
I will save the pointer location and free it so the heap pool for the log
message will be separate and we can avoid this as at that time only a
pointer would need to be read and freed. And we won’t have to assign memory.
I am currently implementing this.

esp_idf/spotflow/src/net/spotflow_mqtt.c
roberthusak: And is it really necessary to make it a one-liner? I usually use the ternary operator `? :` only for short expression, not for a piece of code that involves multiple contained statements. Wouldn't it be simpler to just and `if else` here?

Btw. is it common to use statement expressions in code for ESP-IDF? AFAIK it's not standard C.

noorhaq:
Nope it’s not, more like a personal preference.
I will switch it to the if else block.

esp_idf/spotflow/src/net/spotflow_mqtt.c
```diff
@@ -0,0 +1,150 @@
...
+{
+    while(1)
+    {
+        if(atomic_load(&mqtt_connected))
+        {
+            if(esp_mqtt_client_get_outbox_size(client) == 0)    // Check the mqtt buffer is empty if not 
+            {
+                char *queue_buffer = malloc(CONFIG_SPOTFLOW_CBOR_LOG_MAX_LEN);
+                size_t len = 0;
+                while ((len = queue_read(queue_buffer)) != -1 && atomic_load(&mqtt_connected)) //Check if mqtt disconnect event is not generated.
```
jmasek: boxing int to uint (size_t)

esp_idf/spotflow/src/net/spotflow_mqtt.c
```diff
@@ -0,0 +1,150 @@
...
+ */
+void mqtt_publish(void *pvParameters)
+{
+    while(1)
+    {
+        if(atomic_load(&mqtt_connected))
+        {
+            if(esp_mqtt_client_get_outbox_size(client) == 0)    // Check the mqtt buffer is empty if not 
+            {
+                char *queue_buffer = malloc(CONFIG_SPOTFLOW_CBOR_LOG_MAX_LEN);
```
jmasek: Could you rather allocate a buffer outside the while loop to prevent heap fragmentation - there is no need to do the allocation inside the while cycle.

esp_idf/spotflow/Kconfig.projbuild
noorhaq: Added

esp_idf/spotflow/examples/logs/main/idf_component.yml
noorhaq: Fixed.

esp_idf/spotflow/src/net/spotflow_mqtt.c
noorhaq: Done.

esp_idf/spotflow/src/net/spotflow_mqtt.c
```diff
@@ -0,0 +1,150 @@
...
+ * 
+ * @param pvParameters 
+ */
+void mqtt_publish(void *pvParameters)
+{
+    while(1)
+    {
+        if(atomic_load(&mqtt_connected))
+        {
+            if(esp_mqtt_client_get_outbox_size(client) == 0)    // Check the mqtt buffer is empty if not
```
jmasek, line 129: Do we want to really wait for the buffer to be empty? I am worried that it would mean a performance hit - message would have to wait untill the previous message is sent before enqueuing  - wouldn't something like `CBOR_MAX_LENGTH + outbox_get_size(client->outbox) > client->config->outbox_limit` make more sense? However, default mqtt buffer size seems to be 1024 - so we might need to increase it

esp_idf/spotflow/src/logging/spotflow_log_queue.c
```diff
@@ -0,0 +1,98 @@
...
+}
+
+/**
+ * @brief Initialize the Queue to save the messgaes
+ * 
+ */
+void queue_init(void)
+{
+    struct mq_attr configuration = {
+        .mq_flags = O_NONBLOCK, // ignored by mq_open
```
jmasek: Wouldn't be better to use it in blocking way so the sender thread will wait until new message arrives -> we do not have to do the task delay....

jmasek (changes requested):

esp_idf/spotflow/src/logging/spotflow_log_queue.c
noorhaq: Added xQueue implementation with pointer based refrences.

esp_idf/spotflow/src/logging/spotflow_log_queue.c
noorhaq: No longer valid as using xQueue now.

esp_idf/spotflow/src/logging/spotflow_log_cbor.c
noorhaq: On the len at first the one we are getting is the base buffer and since we don't need that len after the check anymore seems like a waste not to re assign the new updated message length. Like we don't need the len anymore.

esp_idf/spotflow/src/net/spotflow_mqtt.c
noorhaq: No longer valid in new implementation.

esp_idf/spotflow/src/net/spotflow_mqtt.c
noorhaq: In the new configuration it works seperately so at one time we can only have a single outbox message in case of mqtt disconnect or any issues because otherwise it would work without issues.

esp_idf/spotflow/src/logging/spotflow_log_queue.c
noorhaq: No longer valid

jmasek (approved):

roberthusak (approved):

PR #40, opened 2025-11-05
ESP IDF coredump

noorhaq:
Build ID and Coredump needs to be enabled from menuconfig-> components -> Coredump -> enable coredump and save to flash


.github/workflows/esp-idf.yml
```diff
@@ -0,0 +1,103 @@
+name: Build ESP-IDF Project
+
+on:
+  push:
+  pull_request:
```
roberthusak: I recommend keeping either one or the other (probably `push`) so that pushing to a PR doesn't cause launching the action twice.

.github/workflows/esp-idf.yml
```diff
@@ -0,0 +1,103 @@
...
+            exit 1
+          fi
+          echo "✅ Found example path: ${{ matrix.example }}"
+
+      - name: Cache ESP-IDF installation
+        uses: actions/cache@v3
+        with:
+          path: |
+            $HOME/esp-idf
+            $HOME/.espressif
```
roberthusak: Doesn't seem to work - see that it's not present https://github.com/spotflow-io/device-sdk/actions/caches

The latest run shows this warning:
```
Warning: Path Validation Error: Path(s) specified in the action for caching do(es) not exist, hence no cache is being saved.
```

esp_idf/spotflow/examples/Coredump/main/app_main.c
```diff
@@ -0,0 +1,91 @@
...
+#include "driver/gpio.h"
+#include "protocol_examples_common.h"
+
+#include "esp_log.h"
+#include "spotflow.h"
+#include <sys/param.h>
+
+static const char* TAG = "spotflow_testing_coredump";
+
+#define GPIO_INPUT_IO_0     4
```
roberthusak: Is this a button available on ESP32-C6 and ESP32-S3? Or does the developer have to connect their own?

esp_idf/spotflow/examples/Coredump/main/app_main.c
```diff
@@ -0,0 +1,91 @@
...
+
+enum {
+    SPOTFLOW_EVENT_TRIGGER_CRASH
+};
+
+
+// ISR handler — must be IRAM safe if registered with ISR service
+static void IRAM_ATTR button_isr_handler(void* arg)
+{
+    // esp_system_abort("Button pressed -> deliberate crash");
```
roberthusak: Remove commented-out code (other occurrences as well).

esp_idf/spotflow/examples/Coredump/main/Kconfig.projbuild
```diff
@@ -0,0 +1,23 @@
...
+CONFIG_LOG_MAXIMUM_LEVEL_VERBOSE=y
+CONFIG_LOG_MAXIMUM_LEVEL=5
+CONFIG_SPOTFLOW_COREDUMPS_CHUNK_SIZE=1024
+#
+# Core dump
+#
+CONFIG_ESP_COREDUMP_ENABLE_TO_FLASH=y
+# CONFIG_ESP_COREDUMP_ENABLE_TO_UART is not set
+# CONFIG_ESP_COREDUMP_ENABLE_TO_NONE is not set
+# CONFIG_ESP_COREDUMP_DATA_FORMAT_BIN is not set
```
roberthusak: Do we need to keep the unused options in the comments?

esp_idf/spotflow/include/buildid/spotflow_build_id.h
```diff
@@ -0,0 +1,16 @@
...
+
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+int spotflow_build_id_get(const uint8_t** build_id, uint16_t* build_id_len);
+void spotflow_build_id_print(void);
+#ifdef __cplusplus
```
roberthusak, line 13: Nitpick: Add a newline before `#ifdef` for consistency (in other header files as well).

esp_idf/spotflow/include/coredump/spotflow_coredump_queue.h
```diff
@@ -0,0 +1,19 @@
...
+#include "logging/spotflow_log_queue.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+extern bool coredump_found;
+int8_t queue_coredump_push(uint8_t* msg, size_t len);
+bool queue_coredump_read(queue_msg_t* out);
+void queue_coredump_free(queue_msg_t* msg);
+void queue_coredump_init(void);
```
roberthusak: Shouldn't all our non-static symbols be prefixed with `spotflow_` so that they don't collide with user symbols?

esp_idf/spotflow/src/coredump/spotflow_coredump_cbor.c
```diff
@@ -0,0 +1,203 @@
...
+/*not used in current version*/
+#define KEY_OS_VERSION 0x10
+
+#define CORE_DUMP_CHUNK_MESSAGE_TYPE 2
+#define ESP_IDF_OS_VALUE 2
+
+/* Should be approximately 47 bytes (including build ID), putting 64 to be safe */
+#define COREDUMPS_OVERHEAD 64
+
+uint8_t buffer[CONFIG_SPOTFLOW_COREDUMPS_CHUNK_SIZE + COREDUMPS_OVERHEAD];
```
roberthusak: Unused. In Zephyr, it's used to construct the CBOR message and determine its length, then a buffer of this length is allocated and the CBOR message is copied to it. But maybe your approach here is easier - although the result might occupy a slightly larger memory chunk than it needs, the buffer doesn't have to be copied. So maybe just remove `buffer`. @jmasek What do you think?

esp_idf/spotflow/src/coredump/spotflow_coredump_cbor.c
```diff
@@ -0,0 +1,203 @@
...
+	if (err != CborNoError) {
+		SPOTFLOW_LOG("Failed to start CBOR map: %d", err);
+		free(buf);
+		return -1;
+	}
+
+	// Encode key-value pairs one by one
+	err = cbor_encode_uint(&map_encoder, KEY_MESSAGE_TYPE);
+	if (err != CborNoError)
+		goto fail;
```
roberthusak: Add braces `{}` for consistency.

esp_idf/spotflow/src/coredump/spotflow_coredump_cbor.c
```diff
@@ -0,0 +1,203 @@
...
+
+uint8_t buffer[CONFIG_SPOTFLOW_COREDUMPS_CHUNK_SIZE + COREDUMPS_OVERHEAD];
+
+/**
+ * @brief Print Cbor hex for debugging purposes
+ *
+ * @param buf
+ * @param len
+ */
+static void print_cbor_hex(const uint8_t* buf, size_t len)
```
roberthusak: I'd remove this unused (or at least with the commented-out use) function. I imagine that it might have been useful during the development, but I'm not sure it's worth keeping there. @jmasek What do you think?

The same holds for `print_cbor_hex` in `spotflow_log_cbor.h`. If we want to keep those here, shouldn't we maybe introduce an additional Kconfig option instead of having the function call commented out?

esp_idf/spotflow/Kconfig.projbuild
```diff
@@ -65,8 +65,8 @@ menu "Spotflow"
...
         int "Number of messages to retain in case of disconnection"
         default 5
-    
-    config MQTT_BUFFER_SIZE
-        int
+
+    config SPOTFLOW_COREDUMPS_CHUNK_SIZE
+        int 
         default SPOTFLOW_CBOR_LOG_MAX_LEN
 endmenu
```
roberthusak, line 96: Shouldn't be the sending of core dumps using Spotflow configurable in Kconfig? It's opt-in in the Zephyr module.

esp_idf/spotflow/include/logging/spotflow_log_queue.h
```diff
@@ -1,14 +1,22 @@
...
+	uint8_t* ptr;
+	size_t len;
 } queue_msg_t;
 
-void queue_push(uint8_t *msg, size_t len);
-bool queue_read(queue_msg_t *out);
-void queue_free(queue_msg_t *msg);
+void queue_push(uint8_t* msg, size_t len);
+bool queue_read(queue_msg_t* out);
+void queue_free(queue_msg_t* msg);
```
roberthusak: Shouldn't these functions be renamed to `queue_log_*`? Or rather `spotflow_queue_log_*`?

esp_idf/spotflow/spotflow.c
```diff
@@ -16,19 +23,23 @@ vprintf_like_t original_vprintf = NULL;
  */
 void Spotflow_Todo(void)
```
roberthusak: Is this function still needed?

esp_idf/spotflow/src/coredump/spotflow_coredump.c
```diff
@@ -0,0 +1,214 @@
...
+	size_t size;
+	size_t offset;
+	int chunk_ordinal;
+	uint32_t coredump_id;
+} coredump_info_t;
+
+static coredump_info_t coredump_info = { 0 };
+
+/**
+ * @brief If coredump Partion is available return true otherwise false
```
roberthusak, line 32: It also checks whether the coredump is actually stored in the partition, right?

esp_idf/spotflow/src/coredump/spotflow_coredump.c
```diff
@@ -0,0 +1,214 @@
...
+		do {
+			rc = queue_coredump_push(cbor_data, cbor_data_len);
+			if(rc < 0)
+			{
+				vTaskDelay(pdMS_TO_TICKS(50)); //50ms wait
+			}
+			else
+			{
+				vTaskDelay(pdMS_TO_TICKS(50)); //50ms wait //To avoid watchdog trigger
+			}
```
roberthusak: `if` and `else` branches look the same. Btw. why the watchdog could be triggered? Isn't the thread preemptive?

roberthusak (commented):
Looks good, I added mostly minor and stylistic comments.

esp_idf/spotflow/CMakeLists.txt
```diff
@@ -7,8 +8,41 @@ set(SRCS
...
+)
+
+# ---- Optional build ID source ----
+if(CONFIG_SPOTFLOW_GENERATE_BUILD_ID)
+    list(APPEND SRCS "src/buildid/spotflow_build_id.c")
+    list(APPEND PRIV_REQUIRES esp_app_format)
+endif()
+
+# ---- Optional coredump sources ----
+if(CONFIG_ESP_COREDUMP_ENABLE)
```
jmasek, line 28: I do not think this is the correct approach. If the user wants to use our coredump backend -> we should turn on all the necessary ESP components. Not the other way around - if the user activates ESP coredumps, as a side effect, we activate our backend.

esp_idf/spotflow/Kconfig.projbuild
```diff
@@ -65,8 +65,8 @@ menu "Spotflow"
     config SPOTFLOW_MESSAGE_QUEUE_SIZE
         int "Number of messages to retain in case of disconnection"
         default 5
-    
-    config MQTT_BUFFER_SIZE
-        int
+
```
jmasek, line 95: See comment in the cmakelists.txt
Same applies here 

> If the user wants to use our coredump backend -> we should turn on all the necessary ESP components. Not the other way around - if the user activates ESP coredumps, as a side effect, we activate our backend.
I suggest using something like if possible
```
menu "SPOTFLOW Core Dump Settings"

    # Enable ESP-IDF core dump
    config SPOTFLOW_COREDUMP_BACKEND
        bool "Enable automatic core dump settings"
        default n

    if SPOTFLOW_COREDUMP_BACKEND

        # Default destination = Flash
		config ESP_COREDUMP_TO_FLASH
				  bool
				  default y
				
		# Default full .bss/.data + heap dump
		config ESP_COREDUMP_CAPTURE_HEAP_AND_STACK
				  bool
				  default y


    endif

endmenu
```

esp_idf/spotflow/Kconfig.projbuild
```diff
@@ -65,8 +65,8 @@ menu "Spotflow"
...
+    config SPOTFLOW_COREDUMPS_CHUNK_SIZE
```
jmasek: this config options should be available only if spotflow_coredump is enabled

esp_idf/spotflow/Kconfig.projbuild
```diff
@@ -65,8 +65,8 @@ menu "Spotflow"
...
+        int
```
jmasek: A couple more questions regarding the config.
* Is it possible to use only coredump functionality without logs? 
* Shouldn't be the config options more organized so log options are available only when log backend enabled? same for coredump?

esp_idf/spotflow/examples/coredump/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+# The following four lines of boilerplate have to be in your project's CMakeLists
+# in this exact order for cmake to work correctly
+cmake_minimum_required(VERSION 3.16)
+
+
+include($ENV{IDF_PATH}/tools/cmake/project.cmake)
+# "Trim" the build. Include the minimal set of components, main, and anything it depends on.
+idf_build_set_property(MINIMAL_BUILD ON)
```
jmasek, line 8: is there a reason why we are using this? does any functionality depends on it?

esp_idf/spotflow/src/buildid/spotflow_build_id.c
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * @brief To get the Build ID values
+ * 
+ * @param build_id build ID value
+ * @param build_id_len Build ID length
+ * @return int Errors
+ */
+int spotflow_build_id_get(const uint8_t** build_id, uint16_t* build_id_len)
+{
+	 const esp_app_desc_t* app_desc = esp_app_get_description();
```
jmasek, line 20: I see you are using a default esp idf build id! nice. Maybe, do you how exactly it works, what parts are included/omitted in the build id?

esp_idf/spotflow/examples/coredump/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+# The following four lines of boilerplate have to be in your project's CMakeLists
+# in this exact order for cmake to work correctly
+cmake_minimum_required(VERSION 3.16)
```
jmasek, line 3: is there a reason why there is capital C in `Coredumps` folder?

.github/workflows/esp-idf.yml
```diff
@@ -0,0 +1,84 @@
+name: Build ESP-IDF Project
```
jmasek, line 1: Next time, could you please keep features separated in separate PRs so we can merge PRs quicker and comments do not mix together?

esp_idf/spotflow/examples/Coredump/main/app_main.c
jmasek: @noorhaq cannot we use the boot button?

esp_idf/spotflow/examples/Coredump/main/app_main.c
```diff
@@ -0,0 +1,91 @@
...
+{
+	ESP_LOGI(TAG, "[APP] Startup..");
+	ESP_LOGI(TAG, "[APP] Free memory: %" PRIu32 " bytes", esp_get_free_heap_size());
+	ESP_LOGI(TAG, "[APP] IDF version: %s", esp_get_idf_version());
+
+	ESP_ERROR_CHECK(nvs_flash_init());
+	ESP_ERROR_CHECK(esp_netif_init());
+    ESP_ERROR_CHECK(esp_event_loop_create_default());
+
+    // Register our custom event handler
```
jmasek: please add comment what is the event_handler used for

esp_idf/spotflow/examples/Coredump/main/app_main.c
```diff
@@ -0,0 +1,91 @@
...
+
+	/* This helper function configures Wi-Fi or Ethernet, as selected in menuconfig.
+     * Read "Establishing Wi-Fi or Ethernet Connection" section in
+     * examples/protocols/README.md for more information about this function.
+     */
+	ESP_ERROR_CHECK(example_connect());
+
+	spotflow_init();
+	while (1) {
+		// if(atomic_load(&mqtt_connected))
```
jmasek: commented code cleanup - please uncomment or remove. Or add a description if it can be used for a specific purpose.

esp_idf/spotflow/examples/Coredump/main/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+idf_component_register(SRCS "app_main.c"
+                    PRIV_REQUIRES mqtt esp_partition nvs_flash esp_netif app_update esp_driver_gpio
```
jmasek: why we need app_update? because of buildId?

esp_idf/spotflow/examples/Coredump/main/Kconfig.projbuild
```diff
@@ -0,0 +1,23 @@
+CONFIG_EXAMPLE_WIFI_SSID=""
+CONFIG_EXAMPLE_WIFI_PASSWORD=""
```
jmasek: Shouln't be these options rather in sdkonfig.defaults than in KConfig.projbuild?

jmasek (changes requested):

.github/workflows/esp-idf.yml
noorhaq: Done

.github/workflows/esp-idf.yml
noorhaq: Right now removing it will fix the cache in another workflow needs to test it why it's not creating the id in caches.

esp_idf/spotflow/examples/Coredump/main/Kconfig.projbuild
noorhaq: I have captured the default ones like the complete block because if any 2 are active at the same time it will cause issues.

esp_idf/spotflow/src/coredump/spotflow_coredump_cbor.c
noorhaq: Yep this was a oversight. I had planned to use stack then moved to using the heap. For this. So, i am gonna remove it.

esp_idf/spotflow/src/coredump/spotflow_coredump_cbor.c
noorhaq: I think it's worth keeping because if we want to see if anything went wrong with any new boards or configs we can print the raw frames on the cbor side to verify at what point it got corrupted. I will comment it out.

esp_idf/spotflow/Kconfig.projbuild
noorhaq: This one is not exposed. It just makes sure the COREDUMPS_CHUNK_SIZE is equal to SPOTFLOW_CBOR_MAX_LEN to make sure the mqtt buffer can handle them as MQTT buffer is being checked against ```SPOTFLOW_CBOR_LOG_MAX_LEN```

esp_idf/spotflow/spotflow.c
noorhaq: Yep for future add ons and stuff.

esp_idf/spotflow/examples/coredump/CMakeLists.txt
noorhaq: Well it is providing 2 purposes.
One faster builds and smaller size as only components defined in dependencies are build rest are ignored.
Second it makes sure we have included all the components as requirements which are required by our components
https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-guides/build-system.html (Here dp a Ctrl+F search on minimal_build)

esp_idf/spotflow/src/coredump/spotflow_coredump_cbor.c
roberthusak: I'd prefer to have an `#ifdef` with a Kconfig option instead of commenting the code. I think we should avoid commented code in `main` if possible (it' OK for experimenting in PRs though).

esp_idf/spotflow/src/buildid/spotflow_build_id.c
noorhaq: Yes more details can be found here
 https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/app_image_format.html

esp_idf/spotflow/examples/coredump/CMakeLists.txt
noorhaq: Their are examples on the espcoredump components with same name. Either way we can rename it to spotflow_coredump?

esp_idf/spotflow/examples/Coredump/main/app_main.c
noorhaq: Done. Here it was just used for the button event.

esp_idf/spotflow/spotflow.c
jmasek: Can you explain it a bit more? Why won't you add it when needed in the future?

esp_idf/spotflow/spotflow.c
```diff
@@ -16,19 +23,23 @@ vprintf_like_t original_vprintf = NULL;
...
-    original_vprintf = esp_log_set_vprintf(spotflow_log_backend);
+	original_vprintf = esp_log_set_vprintf(spotflow_log_backend);
 
-    Spotflow_Todo(); //Checking for unused set Configs.
-    mqtt_app_start(); // Calling the mqtt_start from the init function.
-    queue_init(); //Initilize the queue
+	Spotflow_Todo(); //Checking for unused set Configs.
+	queue_init(); //Initilize the queue
+	mqtt_app_start(); // Calling the mqtt_start from the init function.
+#ifdef CONFIG_ESP_COREDUMP_ENABLE
```
jmasek, line 30: Please prefer using our custom KConfig option that the default esp one

esp_idf/spotflow/examples/coredump/main/app_main.c
```diff
@@ -0,0 +1,91 @@
...
+
+	ESP_ERROR_CHECK(nvs_flash_init());
+	ESP_ERROR_CHECK(esp_netif_init());
+    ESP_ERROR_CHECK(esp_event_loop_create_default());
+
+    // Register our custom event handler
+    ESP_ERROR_CHECK(esp_event_handler_register(SPOTFLOW_EVENTS, ESP_EVENT_ANY_ID,
+                                               &spotflow_event_handler, NULL));
+
+	 gpio_config_t btn_conf = {
```
jmasek, line 73: the setup is getting quite bulky - could you split the code in the main into multiple functions? Like 
```
setup_and_checks(){
ESP_LOGI(TAG, "[APP] Startup..");
	ESP_LOGI(TAG, "[APP] Free memory: %" PRIu32 " bytes", esp_get_free_heap_size());
	ESP_LOGI(TAG, "[APP] IDF version: %s", esp_get_idf_version());

	ESP_ERROR_CHECK(nvs_flash_init());
	ESP_ERROR_CHECK(esp_netif_init());
    ESP_ERROR_CHECK(esp_event_loop_create_default());
...
}
button_setup(){
  ....
}
```

esp_idf/spotflow/examples/coredump/CMakeLists.txt
jmasek: yes, `spotflow_coredump` and `spotflow_logs` sounds good. btw. where does the conflict comes from? I would expect that the examples are there only for reference and if you are not using both of them simultaneously, it is not problem, or do I miss something?

esp_idf/spotflow/examples/coredump/CMakeLists.txt
jmasek: I see, but my point was that the default is `Off` so I was hesitant to turn it on by default. But I see it is used in esp-idf examples so keep it that way.

esp_idf/spotflow/src/net/spotflow_mqtt.c
```diff
@@ -26,131 +33,167 @@ extern const uint8_t mqtt_spotflow_io_pem_start[]   asm("_binary_x1_root_pem_sta
...
+		    NULL);
 }
 
 /**
  * @brief Seperate Task for publishing the mqtt messages.
  * 
  * @param pvParameters 
  */
-void mqtt_publish(void *pvParameters)
+void mqtt_publish(void* pvParameters)
```
jmasek: It seems that `pvParameters` are not used - do we expect that we will be passing some in the future? does it really make sense to have it there? or is it some convention to rather have it than changing interface?

esp_idf/spotflow/examples/coredump/CMakeLists.txt
noorhaq: Well not their but like say we publish to esp component registery and their it will be a conflict as 2 examples with same name like spotflow/coredump
And espcoredump/coredump. 
But as I write it with the prefix it won't cause confusion.

esp_idf/spotflow/examples/Coredump/main/app_main.c
noorhaq: Moved it to the BOOT button but I still don't wanna connect it to boot.

esp_idf/spotflow/src/coredump/spotflow_coredump.c
noorhaq: Yes, first it checks if the partition exists if yes then it checks whether a valid coredump image exists in the partition or not.

esp_idf/spotflow/src/coredump/spotflow_coredump.c
noorhaq: Yep, kind of a mismatch at my part. It is I was adding a delay because it was retrying and getting a lot of messages on the terminal.

esp_idf/spotflow/Kconfig.projbuild
noorhaq: Done.

esp_idf/spotflow/Kconfig.projbuild
noorhaq: Done.

esp_idf/spotflow/examples/Coredump/main/CMakeLists.txt
noorhaq: Yes and No. TBH for the build id we require the esp_app_format which is added in the ```spotflow/Cmakelist.txt```
But due to some reason without the app_update component it refuses to recognize and compile. ```app_update``` like I tried to debug but had to give up.

esp_idf/spotflow/src/net/spotflow_mqtt.c
noorhaq: It is a convention and the function the xTaskCreate and other Task APIs expect it. Even when we are not passing any like here we are passing NULL.

esp_idf/spotflow/CMakeLists.txt
noorhaq: Ahh kind of confusion here as when we enable the coredump from the menuconfig like using our spotflow configurations it will activate the coredump too. But due to any reason if the coredump is not enabled. And we don't add a check here all the files inside will fail. So that's why I added this one here and I think we shouldn't change it.

esp_idf/spotflow/Kconfig.projbuild
noorhaq: I will create a seperate PR on this. As this will cause conflict issues and require testing with each component separately enabled. Plus configs on examples will change too.

esp_idf/spotflow/src/coredump/spotflow_coredump_queue.c
```diff
@@ -0,0 +1,90 @@
...
+	if (!qmsg.ptr) {
+		SPOTFLOW_LOG("Heap allocation failed");
+		return -1;
+	}
+
+	memcpy(qmsg.ptr, msg, len);
+	// Try to enqueue
+	if (xQueueSend(queue_handle, &qmsg, 0) != pdPASS) {
+        free(qmsg.ptr);
+        return -1; //Could not add wait for it to be freed.
```
jmasek: mixing tabs and spaces. Can you unify it, please?

esp_idf/spotflow/src/net/spotflow_mqtt.c
```diff
@@ -26,131 +33,157 @@ extern const uint8_t mqtt_spotflow_io_pem_start[]   asm("_binary_x1_root_pem_sta
...
-                    SPOTFLOW_LOG("Retrying");
-                   }
-                   else
-                   {
-                    SPOTFLOW_LOG("Message sent successfully. Freeing message in Queue");
-                    queue_free(&msg);
-                   }
+	queue_msg_t msg;
+
+    while (1) {
```
jmasek, line 125: could you please split this main loop to logical parts (functions) - it got quite long.

esp_idf/spotflow/spotflow.c
noorhaq: I have removed it. Like I thought we can have it for consistency. Because we will always have something in todo.

esp_idf/spotflow/examples/coredump/main/app_main.c
noorhaq: It's a example. I have moved the buttons setup to another function but for rest we can keep it as all ESP_IDF based examples use these setups and checks. Like half of this is optional.

noorhaq:
I have addressed most of the changes. let me know if any other changes are required.

esp_idf/spotflow/src/net/spotflow_mqtt.c
```diff
@@ -26,131 +33,157 @@ extern const uint8_t mqtt_spotflow_io_pem_start[]   asm("_binary_x1_root_pem_sta
...
+                        0   // retain
+                    );
+
+                    if (msg_id < 0) {
+                        SPOTFLOW_LOG("Error %d occurred sending MQTT (coredump). Retrying", msg_id);
+                        vTaskDelay(pdMS_TO_TICKS(10)); // Backoff
+                    } else {
+                        SPOTFLOW_LOG("Coredump message sent successfully. Freeing queue entry.");
+                        spotflow_queue_free(&msg);
+						vTaskDelay(pdMS_TO_TICKS(10)); //Give CPU few ticks
```
jmasek: What is the purpose of this?

esp_idf/spotflow/src/net/spotflow_mqtt.c
noorhaq: Yep in the configs one I have. So the next PR will have it addressed. Can we keep it as it is in this one. Otherwise I will have to spend time on merge conflicts and such as in the next one this file has many changes.

esp_idf/spotflow/src/net/spotflow_mqtt.c
noorhaq: So, when the Coredumps messages are sent successfully like all we free the queue allocated to it and free all the memory plus delete the current coredump we have. Because the queue memory for the coredump won't be used until a new one is generated and the device is restarted.

esp_idf/spotflow/src/coredump/spotflow_coredump_queue.c
noorhaq: Done

.github/workflows/esp-idf.yml
```diff
@@ -0,0 +1,88 @@
...
+          sudo apt install -y git wget flex bison gperf python3 python3-pip cmake ninja-build ccache
+          git clone -b v5.5.1 --recursive https://github.com/espressif/esp-idf.git $HOME/esp-idf
+          $HOME/esp-idf/install.sh
+
+      - name: Build example with ESP-IDF
+        run: |
+          source $HOME/esp-idf/export.sh
+          cd "${{ matrix.example }}"
+          idf.py set-target ${{ matrix.target }}
+          idf.py build || true  # continue-on-error
```
jmasek: is this intentional? you won't know if pipeline failed if you do not check the status directly.

jmasek: Could you use `CFLAGS="-Werror"` so we know that the build was successful without warnings?

.github/workflows/esp-idf.yml
```diff
@@ -0,0 +1,88 @@
...
+      - name: Verify example path
+        run: |
+          if [ ! -d "${{ matrix.example }}" ]; then
+            echo "❌ Path does not exist: ${{ matrix.example }}"
+            exit 1
+          fi
+          echo "✅ Found example path: ${{ matrix.example }}"
+
+
+      - name: Install ESP-IDF v5.5.1
```
jmasek: did you consider using official esp-idf docker image so we do not need to install the toolchain everytime? https://hub.docker.com/layers/espressif/idf/release-v5.5/images/sha256-1948c18c8accdd7b94f7e106ab3908a0dfe1898a7acae159ee92e85a70fb6d06

esp_idf/spotflow/examples/coredump/main/app_main.c
```diff
@@ -0,0 +1,117 @@
...
+    #define GPIO_INPUT_IO_0 9     // BOOT button on many C3 devkits
+#elif CONFIG_IDF_TARGET_ESP32C6
+    #define GPIO_INPUT_IO_0 9     // C6 default BOOT button
+#elif CONFIG_IDF_TARGET_ESP32S3
+    #define GPIO_INPUT_IO_0 0     // S3 devkit button (varies)
+#elif CONFIG_IDF_TARGET_ESP32H2
+    #define GPIO_INPUT_IO_0 9     // Example; depends on board
+#else
+    #error "Unsupported target. Please define GPIO_INPUT_IO_0."
+#endif
```
jmasek, line 30: nice, thanks!

.github/workflows/esp-idf.yml
```diff
@@ -0,0 +1,88 @@
...
+          idf.py build || true  # continue-on-error
+
+      - name: Set safe artifact name
+        id: artifact_name
+        run: |
+          EXAMPLE_NAME=$(basename "${{ matrix.example }}")
+          SAFE_NAME="firmware-${{ matrix.target }}-${EXAMPLE_NAME}"
+          echo "name=$SAFE_NAME" >> "$GITHUB_OUTPUT"
+
+      - name: Upload build artifacts
```
jmasek, line 78: Does it make sense to keep built artifacts? Is there any use for them?

esp_idf/spotflow/examples/coredump/main/idf_component.yml
```diff
@@ -0,0 +1,6 @@
+dependencies:
+  protocol_examples_common:
+    path: ${IDF_PATH}/examples/common_components/protocol_examples_common
+  spotflow:
+    override_path: '../../../'
```
jmasek, line 5: should this be changed when released? Btw. our sdk is already in esp-idf component registry?

.github/workflows/esp-idf.yml
noorhaq: I did but on the repo their was some restrictions on using third party runs in the organisational repo. 
So had to go with this custom one. I will make it better like adding caching so it won't build it and download every time just once and have it cached to use for future runs.

.github/workflows/esp-idf.yml
noorhaq: Well kind of yes and no. I am looking into running it as qemu in the runs as well like it will test the scripts to see if the data is being sent as well. So with those the artefacts will be necessary and will be deleted afterwards.

esp_idf/spotflow/examples/coredump/main/idf_component.yml
noorhaq: (I have got the name and got my account approved it requires a few more modifications like small things to make it fully compatible with the component parser and the rest to see and resolve it) in that I was looking and all the examples had it like this like the override_path works a little differently than the path.
And most components had these in examples like this was the recommended way for examples embedded inside the components.
Like separate readmes for each example is required as well.

.github/workflows/esp-idf.yml
noorhaq: I will add this shortly

.github/workflows/esp-idf.yml
noorhaq: https://github.com/espressif/idf-ci I am talking about this one

.github/workflows/esp-idf.yml
jmasek: ok, that would be great, not only build but also run in qemu!
I just checked the esp qemu fork, it looks nice. Keep it there. Should I create an extra milestone for it?

esp_idf/spotflow/examples/coredump/main/idf_component.yml
jmasek: ok, cool.
So it means the default usage of the example is building it and running it from within the library, right? not copying it out to your own folder/project.

esp_idf/spotflow/src/coredump/spotflow_coredump.c
```diff
@@ -0,0 +1,207 @@
...
+ */
+esp_err_t spotflow_coredump_backend(void)
+{
+	size_t coredump_addr = 0;
+	size_t coredump_size = 0;
+
+	// Retrieve the coredump image address and size
+	esp_err_t err = esp_core_dump_image_get(&coredump_addr, &coredump_size);
+	if (err != ESP_OK) {
+		SPOTFLOW_LOG( "No Coredump found.");
```
jmasek, line 76: There can be various errors, but not all of them mean that the coredump was not found, right?

esp_idf/spotflow/src/coredump/spotflow_coredump.c
```diff
@@ -0,0 +1,207 @@
...
+		    (coredump_info.offset + current_chunk_size) >= coredump_info.size;
+		if (is_last_chunk) {
+			SPOTFLOW_LOG( "Processing last chunk of coredump\n");
+		}
+
+		// Get build ID (only for first chunk)
+		const uint8_t* build_id = NULL;
+		uint16_t build_id_len = 0;
+
+#ifdef CONFIG_SPOTFLOW_GENERATE_BUILD_ID
```
jmasek: Maybe `CONFIG_SPOTFLOW_USE_BUILD_ID` would be better - we are not generating it anymore, more like using a generated one.

esp_idf/spotflow/src/coredump/spotflow_coredump.c
```diff
@@ -0,0 +1,207 @@
...
+
+		if (rc < 0) {
+			free(chunk_buffer);
+			SPOTFLOW_LOG( "Failed to encode coredump chunk: %d", rc);
+			return ESP_FAIL;
+		}
+
+		// keep retrying to push data until the mqtt function is able to read and clear it
+		do {
+			rc = spotflow_queue_coredump_push(cbor_data, cbor_data_len);
```
jmasek, line 176: why you are using taskDelay here, when you can use `xQueueSend` with xTicksToWait? is it some kind of optimization?

esp_idf/spotflow/src/net/spotflow_mqtt.c
```diff
@@ -26,131 +33,147 @@ extern const uint8_t mqtt_spotflow_io_pem_start[]   asm("_binary_x1_root_pem_sta
...
+                        SPOTFLOW_LOG("Error %d occurred sending MQTT (log). Retrying", msg_id);
+                        vTaskDelay(pdMS_TO_TICKS(10)); // Backoff
+                    } else {
+                        SPOTFLOW_LOG("Log message sent successfully. Freeing queue entry. \n");
+                        spotflow_queue_free(&msg);
+						vTaskDelay(pdMS_TO_TICKS(10)); //Give CPU few ticks
+                    }
+                }
+                else {
+                    vTaskDelay(pdMS_TO_TICKS(50)); // Sleep if queues empty
```
jmasek, line 174: is there a reason why you are using vTaskDelay rather than xTicksToWait in method `xQueueReceive`

esp_idf/spotflow/src/coredump/spotflow_coredump.c
noorhaq: Yes but we can print the exact error as well. I just didn't see the need for it here. Like other places. Because if it's not able to read the coredump for any reason it won't be able to send. And this coredump is handled by esp idf own execution so we don't have much control over it.

esp_idf/spotflow/src/net/spotflow_mqtt.c
noorhaq: These are just temporary as I mentioned before like the mqtt.c file or net files itself has a whole revamp in the new PR with configs these were basically placeholders so I can have the logic and other things stabilised.

Most of this will be completed in the configs PR and after that their will be a small PR on cleanup and optimisation to reduce memory footprints test with different levels of configuration like build id on and off, logs off coredumps on etc.

esp_idf/spotflow/Kconfig.projbuild
jmasek: ok, added to milestone

esp_idf/spotflow/Kconfig.projbuild
```diff
@@ -1,5 +1,33 @@
...
+            # Default destination = Flash
+            choice ESP_COREDUMP_TO_FLASH_OR_UART
+                default ESP_COREDUMP_ENABLE_TO_FLASH
+            endchoice
+            
+            config ESP_COREDUMP_ENABLE
+                default ESP_COREDUMP_DATA_FORMAT_ELF
+            
+            config ESP_COREDUMP_CAPTURE_DRAM
+                default y
```
jmasek, line 23: why you are not using select/imply?

esp_idf/spotflow/Kconfig.projbuild
```diff
@@ -1,5 +1,33 @@
...
+
+        if SPOTFLOW_COREDUMP_BACKEND
+
+            # Default destination = Flash
+            choice ESP_COREDUMP_TO_FLASH_OR_UART
+                default ESP_COREDUMP_ENABLE_TO_FLASH
+            endchoice
+            
+            config ESP_COREDUMP_ENABLE
+                default ESP_COREDUMP_DATA_FORMAT_ELF
```
jmasek: I do not understand this - how coredump enable is connected to `ESP_COREDUMP_DATA_FORMAT` . I would expect you to select/imply esp_coredump_enable

esp_idf/spotflow/CMakeLists.txt
jmasek: so shouldn't this be `if(SPOTFLOW_COREDUMP_BACKEND)` ?

esp_idf/spotflow/src/buildid/spotflow_build_id.c
jmasek: there are limited information on the page, However, I checked the python and it seems that it does hash from whole file. It is not ideal. Let's keep it that way and we can discuss it in next milestone.

esp_idf/spotflow/src/net/spotflow_mqtt.c
jmasek: ok

.github/workflows/esp-idf.yml
jmasek: what do you mean? I mean using the official docker image that is under Apache Licence 2.0 instead of `ubuntu-latest`

esp_idf/spotflow/src/coredump/spotflow_coredump.c
jmasek: wouldn't be better to print at least the err code? or something that suggest - that maybe we were not able to read coredump or maybe it is missing, or is corrupted, or something else?

esp_idf/spotflow/src/net/spotflow_mqtt.c
jmasek: ok

esp_idf/spotflow/CMakeLists.txt
noorhaq: this is like a dummy check. because we are enabling this config in the Kconfig like the SPOTFLOW_COREDUMP_BACKEND is a wrapper for this and some more. If this fails then everything will fail. We can add duplicate checks.

.github/workflows/esp-idf.yml
noorhaq: Yep now using that one now in the updated job I tried using the official idf-ci image before but that failed, so i just switched to installing.

.github/workflows/esp-idf.yml
noorhaq: Not now. I still have to check if github actions will support for qemu that plus we will need some checks from cloud side like a exposed something to see the data is reached otherwise it would be a dummy check here that the message is sent which will on QOS 0. But has that message been received or not will not be verified.

esp_idf/spotflow/examples/coredump/main/idf_component.yml
noorhaq: yes.

esp_idf/spotflow/src/coredump/spotflow_coredump.c
noorhaq: As you saw. If this fails esp_coredump itself does print this stuff. Like what actually went wrong say size too small and other stuff so it would just be duplicates at this point.

esp_idf/spotflow/src/coredump/spotflow_coredump.c
noorhaq: Nope this will get optimized out in the new mqtt.c as that one is not relying on this. Queue push data it creates a task notify which will trigger the mqtt publish. and same for others. Instead of repeating the same until the mqtt send is able to publish.

esp_idf/spotflow/Kconfig.projbuild
noorhaq: I tried that but the Cmake was giving error as this is choice parameters so select and imply weren't setting the values.

esp_idf/spotflow/Kconfig.projbuild
noorhaq: Oh yeah this one was my mistake fixed it. This one should have had select.

esp_idf/spotflow/Kconfig.projbuild
noorhaq: And as discussed previously all of these will get their respective SPOTFLOW_Wrapper
Like here for instance SPOTFLOW_COREDUMP_CAPTURE_DRAM. default y but users will be able to modify it in menuconfigs.

esp_idf/spotflow/CMakeLists.txt
jmasek: why duplicate check - if `SPOTFLOW_COREDUMP_BACKEND` is not true, there is no reason add these source file to build, right? or do I miss something?

esp_idf/spotflow/src/coredump/spotflow_coredump.c
jmasek: Error `No Coredump found.` is misleading - if you want to print something (that is definitely reasonable), print `unable to get dump image` no misleading information.

noorhaq: Yeah we can change this to something like this. 
Like the thing here is the first one checks if the partition is present or not.
Then it checks for CRC because there will be data inside whether that data is actually a valid coredump or not. 

So, I will change this.

esp_idf/spotflow/CMakeLists.txt
noorhaq: Well yes but we can add like 2 checks because if the spotflow coredump is enabled but at the same time in some config or component they have disabled the ESP_Coredump then we will have linking issues/errors. 
So, in the new Kconfig one we can make it more options and add this in wraps.

jmasek: If someone disables somewhere else `CONFIG_ESP_COREDUMP_ENABLE` -> the solution should be triggering error/warning, not silently omitting our source code!
But disabling `SPOTFLOW_COREDUMP_BACKEND` and therefore omitting our source code is totally fine setup

noorhaq: Exactly that's what being added in the Kconfigs and Cmake right now. 
So these are double checked and all the configs we require are added.

jmasek: ok, so it will be here or in a separate PR?

noorhaq: Seperate PR. Like the next PR has configs options and a MQTT 5 exposed options. Then on a call we can create the Kconfig. As it is minimal just structural changes.

jmasek (dismissed):

jmasek (approved):

PR #46, opened 2025-11-27
Setting log level config from cloud.

noorhaq:


noorhaq:
This has the cloud config settings.
A new mqtt.c file and function which uses events for the mqtt publish. More efficient and while idle it doesn't use CPU cycles. Because when idle the task will be in standby mode waiting for events. 
MQTT publish is created when MQTT is connected and is deleted when MQTT disconnects.

esp_idf/spotflow/Kconfig.projbuild
```diff
@@ -88,9 +82,27 @@ menu "Spotflow"
...
-
+    
     config SPOTFLOW_MESSAGE_QUEUE_SIZE
         int "Number of messages to retain in case of disconnection"
         default 5
 
+    
+    config SPOTFLOW_DEFAULT_SENT_LOG_LEVEL
+        int "Spotflow default sent log level"
+        default 4
```
jmasek: how is default log level done in esp? cannot use default value from esp?

esp_idf/spotflow/Kconfig.projbuild
```diff
@@ -88,9 +82,27 @@ menu "Spotflow"
...
+        range 0 6
```
jmasek: why 6 is maximum?

esp_idf/spotflow/examples/coredump/sdkconfig.defaults
```diff
@@ -7,4 +7,6 @@ CONFIG_SPOTFLOW_DEVICE_ID=""
 CONFIG_SPOTFLOW_INGEST_KEY=""
 CONFIG_PARTITION_TABLE_CUSTOM=y
 CONFIG_PARTITION_TABLE_CUSTOM_FILENAME="partitions.csv"
-CONFIG_PARTITION_TABLE_FILENAME="partitions.csv"
\ No newline at end of file
+CONFIG_PARTITION_TABLE_FILENAME="partitions.csv"
+CONFIG_LOG_MAXIMUM_EQUALS_DEFAULT=n
```
jmasek, line 11: what this mean?

esp_idf/spotflow/Kconfig.projbuild
noorhaq: Okay so it has 0 to 5 like 
0 None, 1 Error, 2 Warning, 3 Info, 4 Error, 5 Verbose, Added 6 in case we want to add critical or any other specific like Spotflow_log_debug here,

esp_idf/spotflow/Kconfig.projbuild
noorhaq: Default value is set to Info https://github.com/espressif/esp-idf/blob/fcae32885b0296b32044cb99ecbdc50d98dddb83/components/log/Kconfig.level#L3C5-L6C13

esp_idf/spotflow/examples/coredump/sdkconfig.defaults
noorhaq: Okay so it was a strange bug, if I don't set this value to n, by default it is y. Any configs I had added and declared in Kconfig were being ignored by the ESP_LOG subsystem.

esp_idf/spotflow/Kconfig.projbuild
jmasek: ok, why the default isn't `LOG_DEFAULT_LEVEL` 
``` 
 config SPOTFLOW_DEFAULT_SENT_LOG_LEVEL
        int "Spotflow default sent log level"
        default LOG_DEFAULT_LEVEL
```

esp_idf/spotflow/examples/coredump/sdkconfig.defaults
jmasek: if it is not working, shouldn't it be in kconfig - like spotflow `depends on !LOG_MAXIMUM_EQUALS_DEFAULT`

esp_idf/spotflow/Kconfig.projbuild
noorhaq: Yeah the default level is info so we can make that change I was testing it too so kept it at a higher debug level. But I can change this to info/default.

esp_idf/spotflow/examples/coredump/sdkconfig.defaults
noorhaq: Nope because it would still work for info, warning and error just not for the debug value like minimal compiled log level will decrease.
Other than that no change would be seen in the overall workflow.

noorhaq: That's why I just disabled it because it was setting a few values and bypassing some which we require at a high level.

esp_idf/spotflow/src/configs/spotflow_config.c
```diff
@@ -0,0 +1,119 @@
+#include <stdbool.h>
+#include <stdint.h>
+
+#include "logging/spotflow_log_backend.h"
+#include "logging/spotflow_log_cbor.h"
+#include "configs/spotflow_config.h"
+#include "configs/spotflow_config_cbor.h"
```
jmasek, line 7: Is there a reason why we use `configs` in plural? other folders have singular - coredump, logging, net, buildId

esp_idf/spotflow/include/configs/spotflow_config_persistance.h
```diff
@@ -0,0 +1,26 @@
...
+#define SPOTFLOW_CONFIG_PERSISTENCE_H
+
+#include <stdbool.h>
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define SPOTFLOW_REPORTED_FLAG_MINIMAL_LOG_SEVERITY (1 << 0)
```
jmasek: This is a mask used to signalize that variable `sent_log_level`, right? wouldn't be better to have name that contains the variable name? rather than minimal_log_severity?

esp_idf/spotflow/src/configs/spotflow_config.c
```diff
@@ -0,0 +1,119 @@
...
+	if ((persisted_settings.flags & SPOTFLOW_REPORTED_FLAG_MINIMAL_LOG_SEVERITY) != 0) {
+		spotflow_config_init_sent_log_level(persisted_settings.sent_log_level);
+	} else {
+		spotflow_config_init_sent_log_level_default();
+	}
+	spotflow_config_init_session();
+}
+
+/**
+ * @brief Initialize the sesson and send a intiail message with current configurations.
```
jmasek: session

esp_idf/spotflow/src/configs/spotflow_config_persistence.c
```diff
@@ -0,0 +1,86 @@
...
+	}
+	ESP_ERROR_CHECK(err);
+}
+
+/**
+ * @brief
+ *
+ * @param settings
+ */
+void spotflow_config_persistence_try_load(struct spotflow_config_persisted_settings* settings)
```
jmasek, line 40: Shouldn't we write to flash only if the value has changed to limit the wear of flash memory?

esp_idf/spotflow/src/configs/spotflow_config.c
```diff
@@ -0,0 +1,119 @@
...
+ *
+ * @param reported_msg
+ */
+static void add_log_severity_to_reported_msg(struct spotflow_config_reported_msg* reported_msg)
+{
+	uint8_t sent_log_level = spotflow_config_get_sent_log_level();
+
+	reported_msg->flags |= SPOTFLOW_REPORTED_FLAG_MINIMAL_LOG_SEVERITY;
+	reported_msg->minimal_log_severity =
+	    spotflow_cbor_convert_log_level_to_severity(sent_log_level);
```
jmasek, line 119: Why are you converting it here and not in the function `spotflow_config_cbor_encode_reported`?

esp_idf/spotflow/src/configs/spotflow_config_cbor.c
```diff
@@ -0,0 +1,181 @@
...
+			return -1;
+		}
+
+		if (key == KEY_MESSAGE_TYPE) {
+			uint64_t msg_type;
+			if (cbor_value_get_uint64(&map_it, &msg_type) != CborNoError) {
+				return -1;
+			}
+			if (msg_type != UPDATE_DESIRED_CONFIGURATION_MESSAGE_TYPE) {
+				return -1;
```
jmasek, line 82: couldyou add log for these errors, please?

esp_idf/spotflow/src/configs/spotflow_config.c
noorhaq: I planned to create a config/Spotflow_config.h so in that we can see which components were enabled and if any are missing like some other component is overwriting any of the configs we have set. 
We can generate a error at this point. So user can fix. 
And I couldn't think of a better name for the folder so I just made it configs. We can change that.

esp_idf/spotflow/include/configs/spotflow_config_persistance.h
noorhaq: I kept the masks to the cbor key bindings like minimal log severity and kind of kept it in line with the zephyr one. Like I kept it in line with zephyr and the other variables set in the struct. To avoid confusion.

esp_idf/spotflow/src/configs/spotflow_config.c
noorhaq: I will fix this.

esp_idf/spotflow/src/configs/spotflow_config.c
noorhaq: It was just a preference to keep the encode clean. Like we save everything the way cloud likes before calling the encode which just encodes the data in struct we provide.

esp_idf/spotflow/src/configs/spotflow_config_cbor.c
noorhaq: I think it would be pointless here as we are already doing a error check on the whole block side. 
We can add these logs to like SPOTFlOW_debug and such and add in like config.

esp_idf/spotflow/src/configs/spotflow_config_persistence.c
```diff
@@ -0,0 +1,86 @@
+#include <stdio.h>
+#include <inttypes.h>
+#include "nvs_flash.h"
+#include "nvs.h"
+#include "esp_err.h"
+#include "configs/spotflow_config_persistance.h"
+#include "logging/spotflow_log_backend.h"
+
+#define SPOTFLOW_STORAGE "Spotflow_S_L"
```
jmasek: is there a reason for different casing between `SPOTFLOW_STORAGE` and `STORAGE_NAMESPACE`

jmasek:
missing changelog

esp_idf/spotflow/src/coredump/spotflow_coredump.c
```diff
@@ -167,39 +168,45 @@ esp_err_t spotflow_coredump_backend(void)
...
 			return ESP_FAIL;
 		}
 
 		// keep retrying to push data until the mqtt function is able to read and clear it
 		do {
 			rc = spotflow_queue_coredump_push(cbor_data, cbor_data_len);
-			vTaskDelay(pdMS_TO_TICKS(50)); //50 ticks wait
-		} while(rc < 0);
+			if (rc < 0) {
+				// vTaskDelay(pdMS_TO_TICKS(500));
```
jmasek: remove if not used.

esp_idf/spotflow/include/net/spotflow_mqtt.h
```diff
@@ -4,15 +4,23 @@
 #include "stdatomic.h"
 #include "mqtt_client.h"
 
+#define SPOTFLOW_MQTT_NOTIFY_COURDUMP (1 << 0) // Flag to trigger sending coredump
```
jmasek: typo coredump

esp_idf/spotflow/src/net/spotflow_mqtt.c
```diff
@@ -4,178 +4,320 @@
...
-                        SPOTFLOW_LOG("Coredump message sent successfully. Freeing queue entry.");
-                        spotflow_queue_coredump_free(&msg);
-                        vTaskDelay(pdMS_TO_TICKS(10)); //Give CPU few ticks
-                    }
-
-                }
-            }
+			// Try to send coredump messages first
+			if (notify_value & SPOTFLOW_MQTT_NOTIFY_COURDUMP) {
+				if(spotflow_coredump_send_message() == 5) {
```
jmasek: magic number 5 - please use a predefined constant instead of a magic number

esp_idf/spotflow/src/net/spotflow_mqtt.c
```diff
@@ -4,178 +4,320 @@
...
+			pdFALSE,	// Do not clear bits
+			pdFALSE,	// wait for ANY bit
+			portMAX_DELAY
+		);
+		EventBits_t clear_mask = 0;
+		SPOTFLOW_LOG("Received notification value: %lu\n", notify_value);
+		// Only proceed if the outbox_size i.e. current message is smaller than overall mqtt_buffer.
+		// 2️⃣ Wait until MQTT outbox has space (zero polling)
+		
+		while (esp_mqtt_client_get_outbox_size(spotflow_client) >=
```
jmasek, line 157: how the default outbox size is defined? not sure if I understand the logic - if occupied space in the buffer is larger than `CONFIG_SPOTFLOW_CBOR_LOG_MAX_LEN`, stop pushing messages to client and wait for messages to be sent? is it correct?

esp_idf/spotflow/src/net/spotflow_mqtt.c
```diff
@@ -4,178 +4,320 @@
...
+/**
+ * @brief
+ *
+ * @param topic
+ * @param data
+ * @param len
+ * @param qos
+ * @return int
+ */
+int spotflow_mqtt_publish_messgae(const char* topic, const uint8_t* data, int len, int qos)
```
jmasek: typo message

esp_idf/spotflow/src/logging/spotflow_log_backend.c
```diff
@@ -125,3 +125,28 @@ int spotflow_log_backend(const char* fmt, va_list args)
...
+	va_list args;
+    va_start(args, level);  // Start processing variadic arguments
+    const char *tag = va_arg(args, const char*);
+    
+    // If tag is NULL (or not provided), apply the level to all log sources
+    if (tag == NULL) {
+		esp_log_level_set("*", level);  // Set log level for all
+    } else {
+        // Otherwise, set the log level for the specified tag
+        esp_log_level_set("*", level);  // Set log level for the provided tag
```
jmasek: shouldn't be the first parameter tag?

jmasek (changes requested):

esp_idf/spotflow/src/net/spotflow_mqtt.c
noorhaq: The mqtt buffer is used by not just our logging system but other services as well. By default it's value in esp idf a buffer size of 6144 bytes. Which you can adjust using the buffer size which we check in Cmakelist.txt if it's smaller than the CBOR_Max_len it forces the mqtt buffer to be equal to it. 
So yes if the buffer has some data which isn't sent and it's greater than the data we are about to add it skips and waits until it clears because we are not clearing the event here. Until the message is sent. 
And if smaller it adds the message to the outbox buffer i.e data being transmitted and moves on by clearing the event and the next things make sure whether the data was sent or not.

esp_idf/spotflow/src/logging/spotflow_log_backend.c
noorhaq: It should be but we are not receiving a tag currently so I have kept it like this. When we start receiving the tag we have to convert it to str and add it here. 
So kind of a placeholder. I will add a comment here.

esp_idf/spotflow/src/configs/spotflow_config.c
roberthusak: I'd also prefer `config` as I understand it as just one configuration with possibly multiple configuration options.

Regarding checking the Kconfig values in `spotflow_config.h`, wouldn't it be cleaner (and possibly faster) to do it in CMake? We already check the MQTT buffer size in `CMakeLists.txt` so we can add other checks as necessary.

esp_idf/spotflow/include/configs/spotflow_config_persistance.h
roberthusak: In Zephyr, `struct spotflow_config_persisted_settings` contains the boolean field `contains_sent_log_level`, so I'd suggest to rename this constant to something like `SPOTFLOW_PERSISTED_SETTINGS_FLAG_SENT_LOG_LEVEL`. Or maybe something less wordy, but it confuses me that the current name references reported properties which are in different module and not directly related to persistence.

esp_idf/spotflow/src/configs/spotflow_config.c
noorhaq: It is but the common workflow is not to check in Cmake but in Kconfig so we can reuse. Like I tried to keep the configs same between our Zephyr and esp idf code.
So in future say we add support for Nuttx, or some other rtos or linux software. And their we might not be using cmake but some other. With the Kconfig.h we can port it easily just copy paste and change or remove the code we don't need. So minimal effort.

esp_idf/spotflow/include/configs/spotflow_config_persistence.h
roberthusak, line 1: Shouldn't we use "persist**e**nce" instead of "persist**a**nce" (for both `.c` and `.h` files)?

esp_idf/spotflow/src/configs/spotflow_config.c
```diff
@@ -0,0 +1,119 @@
...
+{
+	struct spotflow_config_reported_msg reported_msg = {
+		.flags = 0,
+	};
+	add_log_severity_to_reported_msg(&reported_msg);
+
+	int rc = spotflow_config_prepare_pending_message(&reported_msg);
+	if (rc < 0) {
+		SPOTFLOW_LOG(
+		    "Failed to prepare initial reported configuration response message: %d", rc);
```
roberthusak: Please apply this minor fix - it's not a response. My bad, it's in Zephyr as well - can you fix it there as well?
```suggestion
		    "Failed to prepare initial reported configuration message: %d", rc);
```

esp_idf/spotflow/include/configs/spotflow_config.h
```diff
@@ -0,0 +1,18 @@
...
+
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+void spotflow_config_init(void);
+int spotflow_config_init_session(void);
+void spotflow_config_desired_message(const uint8_t* payload, int len);
```
roberthusak: I suggest adding a verb describing what the function does - e.g., `spotflow_config_handle_desired_message`.

esp_idf/spotflow/src/configs/spotflow_config.c
```diff
@@ -0,0 +1,119 @@
...
+/**
+ * @brief Receive the log level from cloud.
+ *
+ * @param payload
+ * @param len
+ */
+void spotflow_config_desired_message(const uint8_t* payload, int len)
+{
+	struct spotflow_config_desired_msg desired_msg;
+	SPOTFLOW_LOG("Deconding Payload\n");
```
roberthusak:
```suggestion
	SPOTFLOW_LOG("Decoding payload\n");
```

esp_idf/spotflow/Kconfig.projbuild
roberthusak: I wouldn't do that - 6 would seem to be even more detailed than verbose which is IMO not needed.

esp_idf/spotflow/Kconfig.projbuild
```diff
@@ -88,9 +82,27 @@ menu "Spotflow"
...
+        range 0 6   
+        help
+            Filters the logs sent to Spotflow cloud according to their severity level.
+            This is the default value that can be dynamically overridden from the cloud.
+            Levels are:
+            - 0 OFF, do not send any logs
+            - 1 ERROR, send LOG_ERR
+            - 2 WARNING, send LOG_WRN in addition to previous level
+            - 3 INFO, send LOG_INF in addition to previous level
+            - 4 DEBUG, send LOG_DBG in addition to previous level
```
roberthusak: These levels should be updated for ESP-IDF.

esp_idf/spotflow/include/configs/spotflow_config_cbor.h
```diff
@@ -0,0 +1,56 @@
...
+	SPOTFLOW_REPORTED_FLAG_ACKED_DESIRED_CONFIG_VERSION = 1 << 2,
+} spotflow_reported_flags_t;
+
+// Flags for desired
+#define SPOTFLOW_DESIRED_FLAG_MINIMAL_LOG_SEVERITY (1 << 0)
+
+// Flags for reported
+#define SPOTFLOW_REPORTED_FLAG_MINIMAL_LOG_SEVERITY (1 << 0)
+#define SPOTFLOW_REPORTED_FLAG_COMPILED_MINIMAL_LOG_SEVERITY (1 << 1)
+#define SPOTFLOW_REPORTED_FLAG_ACKED_DESIRED_CONFIG_VERSION (1 << 2)
```
roberthusak: Why are the flags defined both in the enum and as macros?

esp_idf/spotflow/include/configs/spotflow_config_cbor.h
```diff
@@ -0,0 +1,56 @@
...
+
+// set a flag msg.flags |= SPOTFLOW_REPORTED_FLAG_MINIMAL_LOG_SEVERITY;
+struct spotflow_config_desired_msg {
+	uint8_t flags;
+	uint32_t minimal_log_severity;
+	uint64_t desired_config_version;
+};
+
+// clear a flag msg.flags &= ~SPOTFLOW_REPORTED_FLAG_ACKED_DESIRED_CONFIG_VERSION;
```
roberthusak: Is this comment necessary? I'd a expect a C programmer to know how to clear a flag.

esp_idf/spotflow/src/configs/spotflow_config.c
noorhaq: Okay changed the folder name to config.

noorhaq:
On the configs folder I remember it now when i changed it and rebuilt. During and after build in the build folder esp idf creates a folder called config which has the sdkconfig.h and other things so when I changed it back to config and checked it was causing linking issues because esp idf cmake treats directories with name config a little differently as per Chatgpt. I didn't Investigate it any further. 
@jmasek @roberthusak

esp_idf/spotflow/src/configs/spotflow_config_persistence.c
noorhaq: Fixed it, it was just a convention followed to keep the Namespace in like small case alphabets.

esp_idf/spotflow/src/net/spotflow_mqtt.c
noorhaq: Done.

esp_idf/spotflow/src/configs/spotflow_config_net.c
```diff
@@ -0,0 +1,29 @@
+#include "configs/spotflow_config_net.h"
+#include "net/spotflow_mqtt.h"
+#include "logging/spotflow_log_backend.h"
+
+static uint8_t pending_message_buffer[SPOTFLOW_CONFIG_RESPONSE_MAX_LENGTH];
+static size_t pending_message_length = 0;
+
+int spotflow_config_prepare_pending_message(struct spotflow_config_reported_msg* reported_msg)
+{
+	int rc = spotflow_config_cbor_encode_reported(reported_msg, pending_message_buffer,
```
roberthusak, line 10: Are we sure that there can't be a race condition on the shared buffer? For example, that `spotflow_config_prepare_pending_message` won't be executed while `spotflow_config_send_pending_message` is running?

esp_idf/spotflow/include/configs/spotflow_config_options.h
```diff
@@ -0,0 +1,20 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+uint8_t spotflow_config_get_sent_log_level();
+void spotflow_config_init_sent_log_level_default();
+void spotflow_config_init_sent_log_level(uint8_t level);
+void spotflow_config_set_sent_log_level(uint8_t level);
+void spotflow_config_init_sent_log_level_default();
```
roberthusak: Duplicate?

esp_idf/spotflow/src/configs/spotflow_config_persistence.c
```diff
@@ -0,0 +1,86 @@
...
+#define STORAGE_NAMESPACE "Spotflow"
+/**
+ * @brief
+ *
+ */
+void spotflow_config_persistence_try_init(void)
+{
+	esp_err_t err = nvs_flash_init();
+	if (err == ESP_ERR_NVS_NO_FREE_PAGES || err == ESP_ERR_NVS_NEW_VERSION_FOUND) {
+		ESP_ERROR_CHECK(nvs_flash_erase());
```
roberthusak, line 24: Isn't there a risk that we'll erase some user data unrelated to our module?

esp_idf/spotflow/src/configs/spotflow_config_persistence.c
```diff
@@ -0,0 +1,86 @@
...
+	nvs_handle_t spotflow_handle;
+	esp_err_t err;
+
+	// Open
+	err = nvs_open(STORAGE_NAMESPACE, NVS_READONLY, &spotflow_handle);
+	if (err != ESP_OK) {
+		return;
+	}
+
+	err = nvs_get_u8(spotflow_handle, SPOTFLOW_STORAGE, &settings->sent_log_level);
```
roberthusak: Shouldn't we rename `SPOTFLOW_STORAGE` to `SPOTFLOW_SENT_LOG_LEVEL_KEY` or something like that for better understandability?

esp_idf/spotflow/include/coredump/spotflow_coredump.h
```diff
@@ -11,7 +11,6 @@ extern "C" {
 bool spotflow_is_coredump_available(void);
 esp_err_t spotflow_coredump_backend(void);
 void spotflow_coredump_cleanup(void);
-
```
roberthusak, line 14: Typo? (the newline should be kept)

esp_idf/spotflow/Kconfig.projbuild
```diff
@@ -86,11 +80,33 @@ menu "Spotflow"
             Increase this value if you experience issues with logs being dropped because of buffer overflow.
 
     config SPOTFLOW_DEBUG_MESSAGE_TERMINAL
-        bool "To view the logs on the spotflow side to view any errors while debugging"
+        bool "To view the debug logs on the spotflow side."
```
roberthusak:
```suggestion
        bool "To view the debug logs on the Spotflow side."
```

esp_idf/spotflow/src/logging/spotflow_log_backend.c
```diff
@@ -125,3 +125,28 @@ int spotflow_log_backend(const char* fmt, va_list args)
...
 	return len;
 }
+
+/**
+ * @brief Set the value at low level such that device doesn't generate the logs
+ * 
+ * @param level 
+ * @param ... 
+ */
+void spotflow_log_backend_try_set_runtime_filter(uint8_t level, ...) {
```
roberthusak: Why are variadic arguments used there? Isn't it an overkill? AFAIK the tag isn't currently passed from any call site, so why is the function checking its presence at all? Seems like breaking the YAGNI principle.

esp_idf/spotflow/Kconfig.projbuild
```diff
@@ -86,11 +80,33 @@ menu "Spotflow"
...
+            Levels are:
+            - 0 OFF, do not send any logs
+            - 1 ERROR, send LOG_ERR
+            - 2 WARNING, send LOG_WRN in addition to previous level
+            - 3 INFO, send LOG_INF in addition to previous level
+            - 4 DEBUG, send LOG_DBG in addition to previous level
+
+    config SPOTFLOW_LOG_BACKEND_SET_RUNTIME_FILTERING
+        bool "To enable the logs set"
+        default y
```
roberthusak: Please extend the label and help text so that the developer knows what's actually done when the option is enabled.

Also, if I understand it correctly, `esp_log_level_set` updates the log level not only for logs to the Spotflow module, but to all the log callbacks in the chain - such as when we're sending the logs to the cloud but they are also sent to UART. Therefore, the user might set the log level from the cloud to warning and more verbose messages would stop showing up on UART, which might not be expected. So, I'd prefer to have this option turned off by default. (It's not a problem in Zephyr, because the log level is adjusted there only for the single logging backend.)

esp_idf/spotflow/src/net/spotflow_mqtt.c
```diff
@@ -4,178 +4,320 @@
...
+ * @param topic
+ * @param topic_len
+ * @param data
+ * @param data_len
+ */
+void spotflow_mqtt_on_message(const char* topic, int topic_len, const uint8_t* data, int data_len)
+{
+	SPOTFLOW_LOG("MQTT Message Received on topic: %.*s", topic_len, topic);
+
+	// Compare topic exactly
```
roberthusak: "Exactly" sounds like we are checking the exact match, but we're checking just the prefix (because the actual topic will be longer due to the workspace and device IDs)., so I recommend updating the comment accordingly.

Also, the condition should check the prefix and not just containment - I suppose replacing `!= NULL` by `== topic` should solve that.

esp_idf/spotflow/Kconfig.projbuild
noorhaq: Changed.

esp_idf/spotflow/include/configs/spotflow_config_persistence.h
noorhaq: I used zephyr code as base so the files, function names and arguments same as them. We can change it.

esp_idf/spotflow/include/configs/spotflow_config.h
noorhaq: Again we can do that but we should keep the change in sync with the zephyr code base as well.

esp_idf/spotflow/Kconfig.projbuild
noorhaq: Okay yeah we can do that. On Kconfigs I was planning to go over in a seperate PR. Because that requires a lot of edits and back and forth on each option.

esp_idf/spotflow/src/net/spotflow_mqtt.c
noorhaq: Okay I added the strncmp instead of strstr which should do the following.

esp_idf/spotflow/src/logging/spotflow_log_backend.c
noorhaq: I just kept it here with a variadic argument. So we can keep the design consistent. When we add the tag level controls in the logs we can still use the code base without any changes. I can change it back.

esp_idf/spotflow/src/configs/spotflow_config_net.c
noorhaq: Yes. Because the mqtt task in esp idf is event based. When the event publish is running the incoming data is only saved in net recv buffers and isn't processed. And in case we ever receive it twice the data would be overwritten.

esp_idf/spotflow/Kconfig.projbuild
noorhaq: It's in the Kconfig PR. Which will merge to this PR before merging to main.

esp_idf/spotflow/include/configs/spotflow_config_persistence.h
roberthusak: Using the Zephyr code as a base is okay, the comment addresses just the typo (`a` instead of `e`) in these two file names (`spotflow_config_persistance.h` and `spotflow_config_persistance.c`).

esp_idf/spotflow/include/configs/spotflow_config.h
roberthusak: This function is called `handle_desired_msg` in Zephyr code base. Just adding the prefix `spotflow_config_` because it's not static in ESP-IDF seems consistent to me. And `message` vs. `msg` is IMO just a detail, feel free to use either of them (Zephyr API often uses `msg` so that's why we used it there as well).

esp_idf/spotflow/include/configs/spotflow_config_cbor.h
```diff
@@ -0,0 +1,45 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// Flags for desired
+#define SPOTFLOW_DESIRED_FLAG_MINIMAL_LOG_SEVERITY (1 << 0)
+
+// Flags for reported
+#define SPOTFLOW_PERSISTED_SETTINGS_FLAG_SENT_LOG_LEVEL (1 << 0)
```
roberthusak: Please move this flag to `spotflow_config_persistence.h`, because persistence is unrelated to CBOR encoding. Furthermore, it's now unclear which flags can be used in the field `uint8_t flags` of `struct spotflow_config_persisted_settings`.

esp_idf/spotflow/include/configs/spotflow_config_persistence.h
noorhaq: Yep i will fix this.

esp_idf/spotflow/include/configs/spotflow_config.h
noorhaq: I was just using a naming convention to know where the file is as at the start I wasn't planning to make it static. 
Like if I connect the handle_message to some other event based so that's why at the time i just named it and went past it without editing.
I think here it won't matter much if we use either. So let's go with the current one.

esp_idf/spotflow/include/configs/spotflow_config_cbor.h
noorhaq: Yep I will try to limit it somehow. I am using flags instead of bool to reduce struct padding like here the struct size will be smaller and we can add future expandability with just added flags. 
I will convert it to enums or some other. 
I placed these flags here in cbor instead of the respective files I think in case we add pre compilation Kconfigs to restrict some functionality so we won't have to deal with include errors as cbor will already have them and cbor is checking them. Again not sure i will give it another look.

esp_idf/spotflow/include/configs/spotflow_config.h
roberthusak: It's OK that it isn't static. The original comment just addresses that the name of the function lacks a **verb** - "spotflow_config_desired_message" seems like a name of a structure whereas "spotflow_config_**handle**_desired_message" is a proper function name. So please just rename it.

esp_idf/spotflow/include/configs/spotflow_config_cbor.h
roberthusak: Using flags is OK 👍 I used boolean bitfields in the Zephyr module, but there seem to be some objective reasons to prefer flags instead. My concern isn't with using flags in general but that it should be apparent which flags are supposed be used in which structure. Using enums might help, but even properly named constants should suffice.

I probably see the problem now - why did you remove `SPOTFLOW_REPORTED_FLAG_MINIMAL_LOG_SEVERITY` from this file and replaced it with `SPOTFLOW_PERSISTED_SETTINGS_FLAG_SENT_LOG_LEVEL`? Although they have the same value, they are different things - the former should be used with `spotflow_config_reported_msg.flags` while the latter with `spotflow_config_persisted_settings.flags`.

By using the same value for both there's an artificial coupling between CBOR encoding and configuration persistence, which are two separate things. Therefore, as you wrote, we wouldn't be able to skip compilation of persistence because CBOR encoding code uses this constant.

So, please move the definition of `SPOTFLOW_PERSISTED_SETTINGS_FLAG_SENT_LOG_LEVEL` to `spotflow_config_persistence.h`, re-introduce `SPOTFLOW_REPORTED_FLAG_MINIMAL_LOG_SEVERITY` here, and use each flag with its respective `struct`.

I know it seems like a minor issue, but writing clean code now may save us many headaches in the future 😅

noorhaq: Added using typedef. Do check it out.

jmasek:
> On the configs folder I remember it now when i changed it and rebuilt. During and after build in the build folder esp idf creates a folder called config which has the sdkconfig.h and other things so when I changed it back to config and checked it was causing linking issues because esp idf cmake treats directories with name config a little differently as per Chatgpt. I didn't Investigate it any further. @jmasek @roberthusak

what about folder name `configuration`?

esp_idf/spotflow/include/configs/spotflow_config_options.h
jmasek: not resolved @noorhaq

esp_idf/spotflow/src/configs/spotflow_config_persistence.c
jmasek: not resolved @noorhaq

esp_idf/spotflow/src/configs/spotflow_config_persistence.c
jmasek: not resolved @noorhaq

esp_idf/spotflow/include/configs/spotflow_config_options.h
noorhaq: Removed.

esp_idf/spotflow/src/configs/spotflow_config_persistence.c
noorhaq: We are erasing only if the initialisation fails like if the nvs initialisation fails it means the data inside is already corrupted and we have a option in menuconfig to display a error to user and disable the NVS erase by default. 
They can enable it to erase if init fails but it is disabled by default. 
Because if the init fails meaning all the storage sectors or like the partition table for the nvs is corrupted. Due to some reason.
Here we are not checking our respective storage space like that comes afterwards.

esp_idf/spotflow/src/configs/spotflow_config_persistence.c
noorhaq: Updated.

jmasek (approved):

PR #47, opened 2025-11-27
Added MQTT 5 support.

noorhaq:


esp_idf/spotflow/src/net/spotflow_mqtt.c
```diff
@@ -72,20 +72,23 @@ static void spotflow_mqtt_event_handler(void* handler_args, esp_event_base_t bas
 		break;
 	case MQTT_EVENT_DATA:
 		spotflow_mqtt_handle_data(event);
+#if CONFIG_MQTT_PROTOCOL_5
```
jmasek: This is for expanding, right? Let's remove it and add it when needed

noorhaq: Yes. Should I add a comment or some placeholder or document it somewhere so when we add it in future we know where too.

jmasek: you can add the comment. so it is clear we did not forget something.

noorhaq: Okay.

PR #48, opened 2025-11-28
Add Zephyr compilation to CI

roberthusak:


zephyr/samples/logs/Kconfig
```diff
@@ -12,30 +12,39 @@ choice SPOTFLOW_NETWORK_IFACE
...
-
 config SPOTFLOW_USE_ETH
 	bool "Enable Ethernet for Spotflow samples"
 	imply ETH_DRIVER
 	imply NET_L2_ETHERNET_MGMT
 
 endchoice
 
+# If selected, fulfills SPOTFLOW_SETTINGS dependencies by storing Zephyr settings on flash
+config SPOTFLOW_SAMPLE_CONFIG_PERSISTENCE_FLASH
```
roberthusak, line 34: What do you think about using the `SPOTFLOW_SAMPLE_` prefix for all options defined in Kconfig samples? I think it would be helpful for readers to distinguish sample-specific options from the core Spotflow configuration.

.github/workflows/zephyr.yml
```diff
@@ -0,0 +1,158 @@
+name: Build Zephyr Project
+
+on:
+  push:
```
jmasek, line 4: don't we want to run it only on push to main branch or on branches in PR?

jmasek: consider using this https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#running-your-pull_request-workflow-based-on-files-changed-in-a-pull-request
btw. the current rule I suggest keeping on main branch

zephyr/samples/logs/Kconfig
jmasek: definitely; however, I suggest putting note to prj.conf too

jmasek (changes requested):

zephyr/samples/logs/Kconfig
roberthusak: OK, done 👍

Regarding the prefix, I'll add it to the remaining options later (i.e., in the next PR) together with changes in the Quickstart and docs so that it remains in sync.

.github/workflows/zephyr.yml
roberthusak: You're right, done 👍

jmasek (approved):

PR #50, opened 2025-12-03
WIP: initial implementation

jmasek:


zephyr/src/metrics/spotflow_metrics.c
```diff
@@ -0,0 +1,397 @@
...
+};
+
+static struct spotflow_metric metrics[CONFIG_SPOTFLOW_METRICS_MAX_REGISTERED];
+static struct k_work_delayable metrics_flush_work;
+static bool pipeline_started;
+
+static const char* aggregation_interval_str(void)
+{
+#if CONFIG_SPOTFLOW_METRICS_AGGREGATION_INTERVAL_SECONDS == 0
+	return "PT0S";
```
noorhaq, line 57: Why are we converting interval seconds to string?

zephyr/src/metrics/Kconfig
```diff
@@ -0,0 +1,57 @@
...
+
+config SPOTFLOW_METRICS_NAME_MAX_LEN
+	int "Maximum metric name length"
+	default 32
+
+config SPOTFLOW_METRICS_LABEL_MAX_LEN
+	int "Maximum label key/value length"
+	default 32
+
+config SPOTFLOW_METRICS_AGGREGATION_INTERVAL_SECONDS
```
noorhaq: We have fixed values for this like 0, 60, 3600 and so on. So shouldn't this be a choice so user doesn't add a different value which causes the return value be NULL( in code).

zephyr/src/metrics/spotflow_metrics.c
jmasek: thanks but not ready for review now, still wip

PR #53, opened 2025-12-17
Feature/metrics v2

jmasek:


zephyr/samples/coredumps/prj.conf
```diff
@@ -46,5 +46,3 @@ CONFIG_FLASH_MAP=y
 CONFIG_NVS=y
 CONFIG_SETTINGS=y
 CONFIG_SETTINGS_NVS=y
-
-CONFIG_MAIN_STACK_SIZE=4096
```
jmasek: already configured in spotflow - configdefault

zephyr/samples/logs/prj.conf
```diff
@@ -28,8 +30,6 @@ CONFIG_NET_IPV6=n
 CONFIG_NET_TCP=y
 CONFIG_NET_UDP=y
 
-CONFIG_MAIN_STACK_SIZE=4096
```
jmasek, line 33: already configured in spotflow - configdefault

zephyr/src/coredumps/spotflow_coredumps_net.c
```diff
@@ -5,56 +5,40 @@
...
-
 /**
  * Polls for enqueued coredump chunks and processes them.
 
  * @retval 0 No coredumps were sent.
  * @retval >0 Coredumps were sent.
  * @retval <0 Error occurred while processing coredumps.
  */
-int spotflow_poll_and_process_enqueued_coredump_chunks()
+int spotflow_poll_and_process_enqueued_coredump_chunks(void)
```
jmasek, line 15: simplified logic for getting messages from queue

zephyr/src/logging/spotflow_log_net.c
```diff
@@ -6,49 +6,34 @@
...
-			if (messages_sent_counter == UINT32_MAX) {
-				LOG_INF("Sent %" PRIu32 " messages. Reset.", messages_sent_counter);
-				messages_sent_counter = 0; /* reset counter */
-			}
-		}
-		logs_poll_event.state = K_POLL_STATE_NOT_READY;
+	struct spotflow_mqtt_logs_msg *msg_ptr;
+
+	/* Best-effort: remove message immediately */
+	if (k_msgq_get(&g_spotflow_logs_msgq, &msg_ptr, K_NO_WAIT) != 0) {
```
jmasek: simplified sending

zephyr/src/metrics/spotflow_metrics_aggregator.c
```diff
@@ -0,0 +1,508 @@
...
+	if (ts == NULL) {
+		k_mutex_unlock(&metric->lock);
+		LOG_WRN("Time series pool full for metric '%s' (%u/%u)", metric->name,
+			ctx->timeseries_count, ctx->timeseries_capacity);
+		return -ENOSPC;
+	}
+
+	/* Start aggregation timer on first report (sliding window) */
+	/* Use per-metric flag to prevent race condition with labeled metrics */
+	if (!ctx->timer_started) {
```
jmasek, line 111: I find this a bit controversial. I thought about having a special flag if jitter should be used. let me know what do you think

zephyr/samples/metrics/src/main.c
```diff
@@ -0,0 +1,256 @@
...
+#include "../../wifi-common/wifi.h"
+#endif
+
+LOG_MODULE_REGISTER(metrics_sample, LOG_LEVEL_INF);
+
+/* Uncomment this function to provide your own device ID in runtime */
+/*const char* spotflow_override_device_id()
+{
+	return "my_custom_device_id";
+}*/
```
roberthusak: I suggest we keep this part only in the `logs` sample so that it doesn't draw attention away from metrics usage. The `logs` sample is much simpler, so it shouldn't matter there.

zephyr/samples/metrics/credentials-sample.conf
```diff
@@ -0,0 +1,5 @@
+# Copy this file to credentials.conf and fill in your Spotflow credentials
+# credentials.conf is gitignored to prevent accidental credential leaks
+
+CONFIG_SPOTFLOW_DEVICE_ID="your-device-id"
+CONFIG_SPOTFLOW_INGEST_KEY="your-ingest-key"
```
roberthusak, line 5: Shouldn't we put also Wi-Fi credentials there as in the other samples? (If added here, we should also modify the sample README accordingly.)

zephyr/samples/metrics/src/main.c
```diff
@@ -0,0 +1,256 @@
...
+	LOG_INF("========================================");
+	LOG_INF("Spotflow Metrics Sample Application");
+	LOG_INF("========================================");
+	LOG_INF("");
+	LOG_INF("This sample demonstrates:");
+	LOG_INF("  - System metrics auto-collection");
+	LOG_INF("  - Custom label-less metrics");
+	LOG_INF("  - Custom labeled metrics");
+	LOG_INF("  - Integration with logs");
+	LOG_INF("");
```
roberthusak: I suggest replacing this block by just a simple single message (this block might even look weird when displayed in reverse order in the portal). Additional info about the sample can be in the README or in comments.

zephyr/samples/metrics/src/main.c
```diff
@@ -0,0 +1,256 @@
...
+
+	/* Initialize application metrics */
+	if (init_application_metrics() < 0) {
+		LOG_ERR("Failed to initialize application metrics");
+		return -1;
+	}
+
+	LOG_INF("");
+	LOG_INF("Starting metric reporting...");
+	LOG_INF("");
```
roberthusak: I don't think the empty lines are necessary:

```suggestion
	LOG_INF("Starting metric reporting...");
```

zephyr/samples/metrics/src/main.c
```diff
@@ -0,0 +1,256 @@
...
+	}
+	LOG_INF("Registered metric: app_counter (int, 1MIN)");
+
+	/* Register labeled float metric - aggregated over 1 minute */
+	/* Supports up to 18 unique label combinations (3 endpoints × 2 methods × 3 statuses) */
+	rc = spotflow_register_metric_float_with_labels(
+	    "http_request_duration_ms", SPOTFLOW_AGG_INTERVAL_1MIN,
+	    18, /* max_timeseries: 3 endpoints × 2 methods × 3 statuses = 18 */
+	    3, /* max_labels */
+	    &g_request_duration_metric);
```
roberthusak: I suggest we stick to basic ASCII in C source files:

```suggestion
	/* Supports up to 18 unique label combinations (3 endpoints x 2 methods x 3 statuses) */
	rc = spotflow_register_metric_float_with_labels(
	    "http_request_duration_ms", SPOTFLOW_AGG_INTERVAL_1MIN,
	    18, /* max_timeseries: 3 endpoints x 2 methods x 3 statuses = 18 */
	    3, /* max_labels */
	    &g_request_duration_metric);
```

zephyr/samples/metrics/README.rst
```diff
@@ -0,0 +1,310 @@
...
+  - CPU utilization percentage
+  - MQTT connection state changes
+  - Device reset cause (reported on boot)
+  - Device uptime heartbeat
+
+- **Custom Application Metrics**
+
+  - Label-less metrics: Simple counter and temperature sensor
+  - Labeled metrics: HTTP request duration with endpoint, method, and status labels
+  - Note: Metric names are normalized (lowercase, alphanumeric, underscores only)
```
roberthusak: I don't think the note needs to be here - the sample doesn't show the normalization:

```suggestion
  - Label-less metrics: Simple counter and temperature sensor
  - Labeled metrics: HTTP request duration with endpoint, method, and status labels
```

For the same reason, I'm not sure about keeping the section **Metric Name Normalization** either.

zephyr/samples/metrics/README.rst
```diff
@@ -0,0 +1,310 @@
+.. zephyr:code-sample:: spotflow_metrics
+   :name: Spotflow Metrics
+   :relevant-api: spotflow_metrics
+
+   Demonstrate Spotflow metrics collection and reporting for Zephyr-based IoT devices.
```
roberthusak: Unless we're planning an integration with the Zephyr documentation, this shouldn't be necessary (we shouldn't also use `zephyr-app-commands` below). For our samples to be easily discoverable by Zephyr-related VS Code extensions etc., it should be AFAIK enough to add `sample.yaml` to each sample.

.editorconfig-checker.json
```diff
@@ -1,3 +1,4 @@
 {
-  "SpacesAfterTabs": true
+  "SpacesAfterTabs": true,
+  "Exclude": [".*\\.rst$"]
```
roberthusak: Shouldn't we just allow longer lines in `.editorconfig` as it's now done for Markdown?

CODING_STYLE.md
```diff
@@ -0,0 +1,72 @@
...
+9. **Static function implementations**
+
+### Example
+
+```c
+/*
+ * Copyright (c) 2024 Spotflow
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
```
roberthusak: Do we really need this? Isn't the license file in the repo root sufficient?

If we want to keep it, we should fix the year. We also use [BUSL](https://en.wikipedia.org/wiki/Business_Source_License), not Apache-2.0.

zephyr/src/net/spotflow_session_metadata.c
```diff
@@ -6,26 +6,40 @@
...
 #define SESSION_METADATA_MESSAGE_TYPE 1
 
-/* Expecting at most 26 bytes, adding a safety margin */
-#define MAX_CBOR_SIZE 32
+/* Expecting at most 34 bytes, adding a safety margin */
+#define MAX_CBOR_SIZE 64
 
 LOG_MODULE_DECLARE(spotflow_net, CONFIG_SPOTFLOW_MODULE_DEFAULT_LOG_LEVEL);
 
+static uint64_t device_run_id = 0;
```
roberthusak, line 24: `g_device_run_id`? 🙂

zephyr/src/metrics/system/Kconfig
```diff
@@ -0,0 +1,156 @@
...
+
+	  Total with defaults: ~18 KB
+
+	  To reduce memory, lower SPOTFLOW_METRICS_SYSTEM_STACK_MAX_THREADS.
+
+	  This is added to CONFIG_HEAP_MEM_POOL_ADD_SIZE_SPOTFLOW_METRICS
+	  which is for application metrics.
+
+
+
```
roberthusak: Unnecessary extra lines:

```suggestion

```

zephyr/src/metrics/system/Kconfig
```diff
@@ -0,0 +1,156 @@
...
+config SPOTFLOW_METRICS_SYSTEM_MEMORY
+	bool "Enable memory metrics"
+	default y
+	help
+	  Collect system memory statistics (free/total bytes).
+	  Uses Zephyr sys_mem_stats_get() API.
+
+config SPOTFLOW_METRICS_SYSTEM_HEAP
+	bool "Enable heap metrics"
+	imply SYS_HEAP_RUNTIME_STATS
```
roberthusak: Shouldn't we force it? AFAIK it won't compile without this option.

```suggestion
	select SYS_HEAP_RUNTIME_STATS
```

zephyr/src/metrics/system/Kconfig
```diff
@@ -0,0 +1,156 @@
...
+
+if SPOTFLOW_METRICS_SYSTEM
+
+config SPOTFLOW_METRICS_SYSTEM_MEMORY
+	bool "Enable memory metrics"
+	default y
+	help
+	  Collect system memory statistics (free/total bytes).
+	  Uses Zephyr sys_mem_stats_get() API.
+
```
roberthusak: Duplicate? (not used anywhere)

```suggestion

```

zephyr/src/metrics/system/Kconfig
```diff
@@ -0,0 +1,156 @@
...
+	  MAX_LABELS=4) and number of concurrent timeseries per metric.
+
+	  Per-metric heap = 80 + (230 × max_timeseries)
+
+	  Default estimates (all system metrics enabled):
+	  - Heap metrics (2 metrics × 1 ts): ~0.6 KB
+	  - Network metrics (2 × NETWORK_MAX_INTERFACES=4): ~2 KB
+	  - CPU metric (1 ts): ~0.3 KB
+	  - Connection metric (1 ts): ~0.3 KB
+	  - Stack metrics (2 × STACK_MAX_THREADS=32): ~15 KB (dominates!)
```
roberthusak: I suggest sticking to ASCII here as well:


```suggestion
	  Per-metric heap = 80 + (230 x max_timeseries)

	  Default estimates (all system metrics enabled):
	  - Heap metrics (2 metrics x 1 ts): ~0.6 KB
	  - Network metrics (2 x NETWORK_MAX_INTERFACES=4): ~2 KB
	  - CPU metric (1 ts): ~0.3 KB
	  - Connection metric (1 ts): ~0.3 KB
	  - Stack metrics (2 x STACK_MAX_THREADS=32): ~15 KB (dominates!)
```

zephyr/src/metrics/system/spotflow_reset_helper.h
```diff
@@ -0,0 +1,21 @@
...
+
+#ifndef METRICS_SPOTFLOW_RESET_HELPER_H
+#define METRICS_SPOTFLOW_RESET_HELPER_H
+#include <stddef.h>
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+void report_reboot_reason(void);
```
roberthusak: Missing newline:

```suggestion
#endif

void report_reboot_reason(void);
```

Also, shouldn't the name be `spotflow_reset_report_reason` or similar?

zephyr/src/metrics/system/spotflow_metrics_system_heap.c
```diff
@@ -0,0 +1,80 @@
...
+		LOG_ERR("Failed to get heap stats: %d", ret);
+		return;
+	}
+
+	// TODO: Replace with spotflow_report_metric_uint64 when available
+	int64_t free_bytes_capped =
+	    (heap_stats.free_bytes > INT64_MAX) ? INT64_MAX : (int64_t)heap_stats.free_bytes;
+	int64_t allocated_bytes_capped = (heap_stats.allocated_bytes > INT64_MAX)
+	    ? INT64_MAX
+	    : (int64_t)heap_stats.allocated_bytes;
```
roberthusak: Do we really need to do this (in other metrics as well)? I wouldn't expect an embedded system to have more than 9 exabytes of RAM/traffic 😅

I think it's OK to keep the capping for correctness, but all the potential extra code needed to provide an unsigned variant of integer metrics doesn't seem worth the one extra bit of information.

zephyr/src/net/spotflow_session_metadata.c
```diff
@@ -6,26 +6,40 @@
...
 {
 	const uint8_t* build_id = NULL;
 	uint16_t build_id_len = 0;
 
+	/* Generate device run ID once per boot */
+	if (device_run_id == 0) {
+		uint32_t rand_high = sys_rand32_get();
+		uint32_t rand_low = sys_rand32_get();
+		device_run_id = ((uint64_t)rand_high << 32) | rand_low;
+		LOG_INF("Generated device run ID: %llu", device_run_id);
```
roberthusak: As in the other places in the code:

```suggestion
		LOG_INF("Generated device run ID: %llu", (unsigned long long)device_run_id);
```

zephyr/src/metrics/system/spotflow_metrics_system_stack.c
```diff
@@ -0,0 +1,184 @@
...
+	    max_threads, 1, &g_stack_used_percent_metric);
+	if (rc < 0) {
+		LOG_ERR("Failed to register stack used percent metric: %d", rc);
+		return rc;
+	}
+
+#ifndef CONFIG_SPOTFLOW_METRICS_SYSTEM_STACK_ALL_THREADS
+	k_mutex_init(&g_tracked_threads_mutex);
+	for (int i = 0; i < CONFIG_SPOTFLOW_METRICS_SYSTEM_STACK_MAX_THREADS; i++) {
+		g_tracked_threads[i] = NULL;
```
roberthusak: Is this needed? Global variables (including array contents) are always initialized to zero.

zephyr/src/metrics/system/spotflow_metrics_system_stack.c
```diff
@@ -0,0 +1,184 @@
...
+		LOG_ERR("Stack metrics not initialized");
+		return -EINVAL;
+	}
+
+	k_mutex_lock(&g_tracked_threads_mutex, K_FOREVER);
+
+	for (int i = 0; i < CONFIG_SPOTFLOW_METRICS_SYSTEM_STACK_MAX_THREADS; i++) {
+		if (g_tracked_threads[i] == thread) {
+			k_mutex_unlock(&g_tracked_threads_mutex);
+			return -EINVAL;
```
roberthusak: To unify with metric registration?

```suggestion
			return -EEXIST;
```

zephyr/src/metrics/system/spotflow_metrics_system_stack.c
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	float used_percent = (float)used_bytes / (float)stack_size * 100.0f;
+	rc = spotflow_report_metric_float_with_labels(g_stack_used_percent_metric, used_percent,
+						      labels, 1);
+	if (rc < 0) {
+		LOG_ERR("Failed to report stack used percent metric for %s: %d", thread_label, rc);
+	}
+
+	/* Log using integer format to avoid float-to-double promotion (Zephyr convention) */
```
roberthusak: I haven't found any such convention. However, the support for `%f` must be explicitly enabled, so it's better not to force the user to do it. So maybe something like this?

```suggestion
	/* Use integer format to avoid enabling CONFIG_CBPRINTF_FP_SUPPORT */
```

zephyr/src/metrics/Kconfig
```diff
@@ -0,0 +1,107 @@
...
+	  1. Timeseries size = ~36 + (48 × MAX_LABELS_PER_METRIC) bytes
+	     With MAX_LABELS=4: ~230 bytes per timeseries
+	  2. Number of concurrent timeseries (max_timeseries parameter at registration)
+
+	  Per-metric heap = 80 + (timeseries_size × max_timeseries)
+
+	  Examples (with MAX_LABELS=4, ~230 bytes/ts):
+	  - Dimensionless metric (max_ts=1): 80 + 230 = 310 bytes
+	  - Labeled metric (max_ts=4): 80 + 230×4 = 1KB
+	  - Labeled metric (max_ts=16): 80 + 230×16 = 3.8KB
```
roberthusak: I suggest sticking to ASCII here as well:

```suggestion
	  1. Timeseries size = ~36 + (48 x MAX_LABELS_PER_METRIC) bytes
	     With MAX_LABELS=4: ~230 bytes per timeseries
	  2. Number of concurrent timeseries (max_timeseries parameter at registration)

	  Per-metric heap = 80 + (timeseries_size x max_timeseries)

	  Examples (with MAX_LABELS=4, ~230 bytes/ts):
	  - Dimensionless metric (max_ts=1): 80 + 230 = 310 bytes
	  - Labeled metric (max_ts=4): 80 + 230x4 = 1KB
	  - Labeled metric (max_ts=16): 80 + 230x16 = 3.8KB
```

zephyr/src/metrics/spotflow_metrics_types.h
```diff
@@ -0,0 +1,161 @@
...
+#include <zephyr/kernel.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @brief Aggregation interval enumeration
+ *
+ * Maps to ISO 8601 duration strings as specified in the ingestion protocol.
```
roberthusak: Maybe an outdated comment? Those duration strings are now only in the comments. The values correspond to those in the ingestion protocol though.

zephyr/src/metrics/spotflow_metrics_types.h
```diff
@@ -0,0 +1,161 @@
...
+	const char* key;
+	const char* value;
+};
+
+/**
+ * @brief Internal label storage (copied strings)
+ *
+ * Sizes per design specification:
+ * - Key: max 16 characters (SPOTFLOW_MAX_LABEL_KEY_LEN)
+ * - Value: max 32 characters (SPOTFLOW_MAX_LABEL_VALUE_LEN)
```
roberthusak: Is this comment necessary?

CODING_STYLE.md
```diff
@@ -0,0 +1,72 @@
...
+   - Project headers (e.g., `#include "metrics/spotflow_metrics_backend.h"`)
+   - Zephyr/system headers (e.g., `#include <zephyr/kernel.h>`)
+3. **LOG_MODULE_REGISTER / LOG_MODULE_DECLARE**
+4. **Defines and macros**
+5. **Type definitions** (structs, enums, typedefs)
+6. **Static variables**
+7. **Forward declarations of static functions**
+8. **Public function implementations**
+9. **Static function implementations**
+
```
roberthusak, line 25: Should we mention also the usage of Doxygen comments?

zephyr/src/logging/spotflow_log_net.h
```diff
@@ -5,8 +5,7 @@
 extern "C" {
 #endif
 
-void init_logs_polling();
-int poll_and_process_enqueued_logs();
+int poll_and_process_enqueued_logs(void);
```
roberthusak: `spotflow_poll_and_process_enqueued_logs`? (seems to have slipped through the cracks earlier)

zephyr/src/metrics/spotflow_metrics_net.h
```diff
@@ -0,0 +1,44 @@
...
+ * @brief Poll and process one enqueued metric message
+ *
+ * Dequeues one message from the metrics queue and publishes via MQTT.
+ * Called repeatedly by processor thread to drain the queue.
+ *
+ * Memory Ownership: Processor thread ALWAYS frees message memory
+ * (success or failure) after dequeue completes.
+ *
+ * MQTT Retry: Retries infinitely on -EAGAIN to preserve message ordering.
+ * Does NOT requeue (would break ordering).
```
roberthusak: These comments seem wrong. The function removes the message from the queue only after a successful publish.

Also, the MQTT retry logic is implemented by callers (i.e., not by `spotflow_poll_and_process_enqueued_metrics`), so I don't think it should be described here.

zephyr/src/metrics/spotflow_metrics_cbor.h
```diff
@@ -0,0 +1,68 @@
...
+ * @brief Encode metric message to CBOR format
+ *
+ * Allocates and returns a CBOR-encoded buffer. Caller owns the buffer
+ * and is responsible for freeing it with k_free().
+ *
+ * @param metric Metric base handle
+ * @param ts Time series state to encode
+ * @param timestamp_ms Device uptime in milliseconds when aggregation window closed
+ * @param sequence_number Sequence number for this message (caller must increment atomically)
+ * @param cbor_data Output: allocated CBOR buffer (caller must free if enqueue fails)
```
roberthusak: I'd remove notes specific to the current usage of the function: (i.e. not about the function itself)

```suggestion
 * @param sequence_number Sequence number for this message
 * @param cbor_data Output: allocated CBOR buffer
```

zephyr/src/metrics/spotflow_metrics_cbor.c
```diff
@@ -0,0 +1,355 @@
...
+				 struct metric_timeseries_state* ts, int64_t timestamp_ms,
+				 uint64_t sequence_number, uint8_t** cbor_data, size_t* cbor_len)
+{
+	if (metric == NULL || ts == NULL || cbor_data == NULL || cbor_len == NULL) {
+		return -EINVAL;
+	}
+
+	if (metric->agg_interval == SPOTFLOW_AGG_INTERVAL_NONE) {
+		LOG_ERR("This function should not be used for non-aggregated metrics");
+		return -EINVAL; /* This function is for non-aggregated metrics only */
```
roberthusak: The comment contradicts the error message.

```suggestion
		return -EINVAL;
```

zephyr/src/metrics/spotflow_metrics_cbor.c
```diff
@@ -0,0 +1,355 @@
...
+
+		for (uint8_t i = 0; i < label_count && succ; i++) {
+			succ = succ &&
+			    zcbor_tstr_put_term(state, labels[i].key, SPOTFLOW_MAX_LABEL_KEY_LEN);
+			succ = succ &&
+			    zcbor_tstr_put_term(state, labels[i].value,
+						SPOTFLOW_MAX_LABEL_VALUE_LEN);
+		}
+
+		succ = succ && zcbor_map_end_encode(state, label_count);
```
roberthusak, line 154: Shouldn't we call `encode_labels` instead?

zephyr/src/metrics/spotflow_metrics_cbor.c
roberthusak: Also, shouldn't the function name reflect this requirement?

zephyr/src/metrics/spotflow_metrics_registry.c
```diff
@@ -0,0 +1,347 @@
...
+ *
+ * IMPORTANT: Both int_metric and float_metric have identical memory layout
+ * (single 'base' member as first field), allowing safe access via int_metric.base
+ * regardless of actual type. This pattern is used throughout registry operations
+ * to simplify slot management.
+ */
+union metric_registry_entry {
+	struct spotflow_metric_int int_metric;
+	struct spotflow_metric_float float_metric;
+};
```
roberthusak: Can't this union be replaced by `spotflow_metric_base`? `float_metric` isn't used at all and `int_metric` is used only to get to the `base` field.

zephyr/src/metrics/spotflow_metrics_backend.h
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * Events are point-in-time occurrences that are not aggregated over time.
+ * This function is equivalent to calling spotflow_report_metric_int(metric, 1)
+ * with PT0S aggregation interval. The value is always 1 (event occurred).
+ *
+ * @param metric Metric handle from registration (must be label-less)
+ *
+ * @return 0 on success, negative errno on failure
+ *         -EINVAL: Invalid metric handle or metric is labeled
+ *         -ENOMEM: Metric queue is full
```
roberthusak: `-EAGAIN` can occur here (and in the next function) as well, right?

zephyr/src/metrics/spotflow_metrics_backend.h
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ * @brief Report a label-less integer metric value
+ *
+ * @param metric Metric handle from registration
+ * @param value Integer value to report
+ *
+ * @return 0 on success, negative errno on failure
+ *         -EINVAL: Invalid metric handle
+ *         -EAGAIN: Aggregator busy (rare, retry)
```
roberthusak: The actual reason would be that the metrics message queue is full, not "Aggregator busy", right?

Btw. I suggest mentioning the error codes (or at least the most interesting ones) in the comments of other functions in this file as well.

zephyr/src/metrics/spotflow_metrics_backend.c
```diff
@@ -0,0 +1,185 @@
...
+			LOG_ERR("Label key or value is NULL at index %u", i);
+			return -EINVAL;
+		}
+		if (strlen(labels[i].key) >= SPOTFLOW_MAX_LABEL_KEY_LEN) {
+			LOG_WRN("Label key at index %u will be truncated", i);
+		}
+		if (strlen(labels[i].value) >= SPOTFLOW_MAX_LABEL_VALUE_LEN) {
+			LOG_WRN("Label value at index %u will be truncated", i);
+		}
+	}
```
roberthusak: I suggest moving this code (and maybe also the `if` above it) to a helper function - it's already used at 3 places.

zephyr/src/metrics/spotflow_metrics_cbor.c
```diff
@@ -0,0 +1,355 @@
...
+ */
+
+#include "spotflow_metrics_cbor.h"
+
+#include <zcbor_common.h>
+#include <zcbor_encode.h>
+#include <zephyr/kernel.h>
+#include <zephyr/logging/log.h>
+
+LOG_MODULE_REGISTER(spotflow_metrics_cbor, CONFIG_SPOTFLOW_METRICS_PROCESSING_LOG_LEVEL);
```
roberthusak: Shouldn't we keep just a single module `spotflow_metrics` (now it seems that every `.c` file has its own module), as in the case of logs?

zephyr/src/metrics/spotflow_metrics_aggregator.c
```diff
@@ -0,0 +1,525 @@
...
+#include <zephyr/logging/log.h>
+#include <zephyr/random/random.h>
+#include <string.h>
+#include <limits.h>
+#include <float.h>
+
+LOG_MODULE_REGISTER(spotflow_metrics_agg, CONFIG_SPOTFLOW_METRICS_PROCESSING_LOG_LEVEL);
+
+/* External message queue (defined in spotflow_metrics_net.c) */
+extern struct k_msgq g_spotflow_metrics_msgq;
```
roberthusak: I recommend putting this declaration to `spotflow_metrics_net.h` to better show the dependency on it. Logs and coredumps do it the same way.

zephyr/src/metrics/spotflow_metrics_aggregator.c
```diff
@@ -0,0 +1,525 @@
...
+	/* this to safely rollback metric registration on failure. */
+
+	/* Allocate aggregator context */
+	struct metric_aggregator_context* ctx = k_malloc(sizeof(*ctx));
+	if (!ctx) {
+		return -ENOMEM;
+	}
+
+	/* Allocate time series array */
+	ctx->timeseries = k_calloc(metric->max_timeseries, sizeof(struct metric_timeseries_state));
```
roberthusak: I think `k_malloc` and `k_calloc` speak for themselves:

```suggestion
	struct metric_aggregator_context* ctx = k_malloc(sizeof(*ctx));
	if (!ctx) {
		return -ENOMEM;
	}

	ctx->timeseries = k_calloc(metric->max_timeseries, sizeof(struct metric_timeseries_state));
```

zephyr/src/metrics/spotflow_metrics_aggregator.c
```diff
@@ -0,0 +1,525 @@
...
+
+	if (metric->agg_interval == SPOTFLOW_AGG_INTERVAL_NONE) {
+		int rc = flush_no_aggregation_metric(metric, labels, label_count, value_int,
+						     value_float);
+		k_mutex_unlock(&metric->lock);
+		return rc;
+	}
+
+	/* Find or create time series */
+	struct metric_timeseries_state* ts = find_or_create_timeseries(ctx, labels, label_count);
```
roberthusak:
```suggestion
	struct metric_timeseries_state* ts = find_or_create_timeseries(ctx, labels, label_count);
```

zephyr/src/metrics/spotflow_metrics_aggregator.c
roberthusak: Yeah, I agree that it would be better if the user can disable the jitter in Kconfig. Do I understand correctly that the jitter isn't necessary if `SPOTFLOW_METRICS_QUEUE_SIZE` is large enough? So on SoCs with large enough SRAM (and CPU for handling bursts) the jitter can be disabled?

zephyr/src/metrics/spotflow_metrics_aggregator.c
```diff
@@ -0,0 +1,525 @@
...
+	struct spotflow_metric_base* metric = ctx->metric;
+
+	/* Capture timestamp when aggregation window closes */
+	int64_t timestamp_ms = k_uptime_get();
+	LOG_DBG("Aggregation window closed for metric '%s' at %lld ms", metric->name, timestamp_ms);
+
+	k_mutex_lock(&metric->lock, K_FOREVER);
+
+	LOG_DBG("Aggregation timer expired for metric '%s' (%u active time series)", metric->name,
+		ctx->timeseries_count);
```
roberthusak: Shouldn't just one of these debug messages be kept?

zephyr/src/metrics/spotflow_metrics_aggregator.c
```diff
@@ -0,0 +1,525 @@
...
+	} else {
+		LOG_ERR("Invalid metric type: %d", type);
+	}
+}
+
+/**
+ * @brief Find or create time series slot
+ *
+ * Uses direct string comparison for label matching. O(n) linear search
+ * is acceptable since max_timeseries ≤ 256 and labels are small.
```
roberthusak:
```suggestion
 * is acceptable since max_timeseries <= 256 and labels are small.
```

zephyr/src/coredumps/spotflow_coredumps_net.c
roberthusak: LGTM 👍

zephyr/src/logging/spotflow_log_net.c
roberthusak: LGTM 👍

roberthusak (commented):
The PR looks good, most comments added by me are nitpicks.

zephyr/samples/metrics/src/main.c
```diff
@@ -0,0 +1,256 @@
...
+
+	if (rc < 0) {
+		LOG_ERR("Failed to report request duration: %d", rc);
+	} else {
+		LOG_DBG("Reported request: %s %s -> %s (%.2f ms)", method, endpoint, status,
+			duration_ms);
+	}
+}
+
+#ifdef CONFIG_SPOTFLOW_USE_ETH
```
noorhaq, line 221: I am not sure why we have the ETH stuff here in sample. Can't we move it to the wifi common and call it net common(or something similar) others so, the example focuses on the metrics? Plus we are already using wifi file so we can clean it up a little.

zephyr/samples/metrics/src/main.c
```diff
@@ -0,0 +1,256 @@
...
+					  &g_app_counter_metric);
+	if (rc < 0) {
+		LOG_ERR("Failed to register app_counter metric: %d", rc);
+		return rc;
+	}
+	LOG_INF("Registered metric: app_counter (int, 1MIN)");
+
+	/* Register labeled float metric - aggregated over 1 minute */
+	/* Supports up to 18 unique label combinations (3 endpoints × 2 methods × 3 statuses) */
+	rc = spotflow_register_metric_float_with_labels(
```
noorhaq, line 119: Instead of both using SPOTFLOW_AGG_INTERVAL_1MIN can we switch to like numbers so it shows on one we can use custome time frame as well.

zephyr/samples/metrics/src/main.c
```diff
@@ -0,0 +1,256 @@
...
+	connect_to_wifi();
+#endif
+
+#ifdef CONFIG_SPOTFLOW_USE_ETH
+	LOG_INF("Initializing Ethernet...");
+	turn_on_dhcp_when_device_is_up();
+#endif
+
+	/* Initialize application metrics */
+	if (init_application_metrics() < 0) {
```
noorhaq: Can we use rc = init_application_metrics()
Just for like symmetry like the rest of the code for error handling.

zephyr/src/coredumps/spotflow_coredumps_net.c
```diff
@@ -5,56 +5,45 @@
...
+
+	/* Peek without removing - returns non-zero if queue empty */
+	if (k_msgq_peek(&g_spotflow_core_dumps_msgq, &msg_ptr) != 0) {
+		return 0; /* Queue empty */
+	}
+
+	/* Publish while message is still safely in queue */
+	int rc = spotflow_mqtt_publish_ingest_cbor_msg(msg_ptr->payload, msg_ptr->len);
+	if (rc == -EAGAIN) {
+		/* Temporary, retry later without aborting connection */
```
noorhaq, line 27: Can we add a little delay here like few cpu cycles so it checks after it. So those cycles can be used in other processes.
Because if it checks in the next cpu cycle the resource might still not be free. Just a thought. Even without it would work.

zephyr/src/metrics/system/Kconfig
```diff
@@ -0,0 +1,156 @@
...
+	  Collect CPU utilization percentage using Zephyr's cpu_load module.
+
+config SPOTFLOW_METRICS_SYSTEM_CONNECTION
+	bool "Enable connection state metrics"
+	default y
+	help
+	  Report MQTT connection state changes (connected=1, disconnected=0).
+	  Event-driven metric (not periodic).
+
+config SPOTFLOW_METRICS_SYSTEM_RESET_CAUSE
```
noorhaq, line 54: Won't it be like a dublicate as we might be sending the coredump too after this. So we can document it or force it turn off or on depending on if coredumo is enabled.

zephyr/src/metrics/system/spotflow_metrics_system.h
```diff
@@ -0,0 +1,84 @@
...
+#if CONFIG_SPOTFLOW_METRICS_SYSTEM_AGGREGATION_INTERVAL == 0
+#define SPOTFLOW_METRICS_SYSTEM_AGG_INTERVAL SPOTFLOW_AGG_INTERVAL_NONE
+#elif CONFIG_SPOTFLOW_METRICS_SYSTEM_AGGREGATION_INTERVAL == 60
+#define SPOTFLOW_METRICS_SYSTEM_AGG_INTERVAL SPOTFLOW_AGG_INTERVAL_1MIN
+#elif CONFIG_SPOTFLOW_METRICS_SYSTEM_AGGREGATION_INTERVAL == 3600
+#define SPOTFLOW_METRICS_SYSTEM_AGG_INTERVAL SPOTFLOW_AGG_INTERVAL_1HOUR
+#elif CONFIG_SPOTFLOW_METRICS_SYSTEM_AGGREGATION_INTERVAL == 86400
+#define SPOTFLOW_METRICS_SYSTEM_AGG_INTERVAL SPOTFLOW_AGG_INTERVAL_1DAY
+#else
+#error "Invalid SPOTFLOW_METRICS_SYSTEM_AGGREGATION_INTERVAL: must be 0, 60, 3600, or 86400"
```
noorhaq, line 18: Can't we add here like say, custom so a custom time interval can be added. as well like weekly or daily reporting or something. Instead of being bound by our params.

zephyr/src/metrics/system/spotflow_reset_helper.c
```diff
@@ -0,0 +1,107 @@
...
+	/* Report as immediate event metric */
+	static struct spotflow_metric_int* reset_cause_metric;
+	rc = spotflow_register_metric_int_with_labels(
+	    SPOTFLOW_METRIC_NAME_BOOT_RESET, SPOTFLOW_AGG_INTERVAL_NONE, 1, 1, &reset_cause_metric);
+	if (rc < 0) {
+		LOG_ERR("Failed to register reset cause metric: %d", rc);
+		return;
+	}
+
+	reset_cause_to_string(cause, reset_str, sizeof(reset_str));
```
noorhaq, line 57: If we have a error cause integer. Can't we map it on the cloud side instead of adding more overhead on devices with very limited memory. 🤔

zephyr/src/metrics/system/spotflow_reset_helper.c
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	buf[0] = '\0';
+
+	if (cause == 0U) {
+		snprintk(buf, buf_len, "UNKNOWN");
+		return;
+	}
+
+	for (size_t i = 0; i < RESET_CAUSE_COUNT; i++) {
```
noorhaq, line 88: What is this RESET CAUSE COUNT doing? Like we already have it mapped above.

zephyr/src/metrics/spotflow_metrics_aggregator.c
```diff
@@ -0,0 +1,525 @@
...
+LOG_MODULE_REGISTER(spotflow_metrics_agg, CONFIG_SPOTFLOW_METRICS_PROCESSING_LOG_LEVEL);
+
+/* External message queue (defined in spotflow_metrics_net.c) */
+extern struct k_msgq g_spotflow_metrics_msgq;
+
+static bool labels_equal(const struct metric_timeseries_state* ts,
+			 const struct spotflow_label* labels, uint8_t label_count);
+static void update_aggregation_int(struct metric_timeseries_state* ts, int64_t value);
+static void update_aggregation_float(struct metric_timeseries_state* ts, float value);
+static int64_t get_interval_ms(enum spotflow_agg_interval interval);
```
noorhaq: Can we have it unsigned and save memory like interval will always be positive .

zephyr/src/metrics/spotflow_metrics_backend.h
```diff
@@ -0,0 +1,108 @@
...
+ */
+int spotflow_report_metric_int(struct spotflow_metric_int* metric, int64_t value);
```
noorhaq, line 22: We have to add support for unisgned values as well. Like uint8_t, int8_t, and the same for 16 and 32 bytes. Becasue we mostly have defined sized variables and with this during mapping it can cause issues. Between unisigned and signed variables. And memory being wasted too.

noorhaq: On the cloud side, we won't have to change or add this functionality as for it, it won't cause any issues.

Specially when in future we are saving logs say in file for devices with limited connectivity.

zephyr/src/metrics/spotflow_metrics_cbor.c
```diff
@@ -0,0 +1,355 @@
...
+static bool encode_labels(zcbor_state_t* state, const struct metric_label_storage* labels,
+			  uint8_t label_count);
+static void encode_metric_header(struct spotflow_metric_base* metric, int64_t timestamp_ms,
+				 uint64_t sequence_number, zcbor_state_t state[3], bool* succ);
+static bool encode_aggregation_stats(zcbor_state_t* state, struct spotflow_metric_base* metric,
+				     struct metric_timeseries_state* ts);
+static int finalize_cbor_output(uint8_t* buffer, zcbor_state_t* state, uint8_t** cbor_data,
+				size_t* cbor_len);
+
+int spotflow_metrics_cbor_encode(struct spotflow_metric_base* metric,
```
noorhaq: We have to make some changes here so we have a single cbor file for all, like logs, coredump, metrics etc. 
As this is increasing code footprint like first actual protocol then the cbor conversions. So with more modules in future, this would grow into a full-fledged application instead of bring a low size add on whcih can provide support. 
Like this should be in a ToDo or something. Because each cbor then comes with it's own temporary buffers and all.

zephyr/src/metrics/spotflow_metrics_cbor.c
```diff
@@ -0,0 +1,355 @@
...
+
+	/* Start CBOR map with 4 entries */
+	succ = succ && zcbor_map_start_encode(state, 4);
+
+	/* messageType = 5 */
+	succ = succ && zcbor_uint32_put(state, KEY_MESSAGE_TYPE);
+	succ = succ && zcbor_uint32_put(state, METRIC_MESSAGE_TYPE);
+
+	/* metricName = "uptime_ms" */
+	succ = succ && zcbor_uint32_put(state, KEY_METRIC_NAME);
```
noorhaq, line 204: We should then remove these metrics from log file. So we are not sending dublicate data.

zephyr/src/metrics/spotflow_metrics_net.h
noorhaq: We need to unify some stuff here. Like if each component will come with it's own queue aren't we wasting space. Like say with default values a size 3 queue is generated of size 2048 but with max time 96400s to reduce the frequency of polling aren't we just hogging the space which our logs can use. 
Because this during init this memory would be reserved for metrics.

zephyr/src/metrics/Kconfig
```diff
@@ -0,0 +1,107 @@
...
+
+	  Default 8KB supports ~25 dimensionless metrics or ~8 labeled metrics
+	  with 4 timeseries each. Adjust based on your application needs.
+
+config SPOTFLOW_METRICS_PROCESSING_LOG_LEVEL
+	int "Metrics subsystem log level"
+	default SPOTFLOW_MODULE_DEFAULT_LOG_LEVEL
```
noorhaq, line 81: Okay this was what I was talking about in chat. Can we in the zephyr/Kconfig
Add some more info here. Because when I was glancing over it previously on first look it made me think it is the overall log level. And not the spotflow subsystem log level. 

`config SPOTFLOW_MODULE_DEFAULT_LOG_LEVEL
	int "Spotflow SDK backend level"
	default 2
	range 0 4
	help
		Sets default log level for all Spotflow components only. Not the overall log levels.
		Defaults to 2 (WARNING).
		Levels are:
		- 0 OFF, do not write
		- 1 ERROR, only write LOG_ERR
		- 2 WARNING, write LOG_WRN in addition to previous level
		- 3 INFO, write LOG_INF in addition to previous level
		- 4 DEBUG, write LOG_DBG in addition to previous level`

.gitignore
```diff
@@ -13,3 +13,4 @@ modules/*west
 **/dependencies.lock
 **/.clangd
 esp_idf/spotflow/dist/*
+.claude/
```
noorhaq: Do we need this? In .gitignore.

zephyr/src/net/spotflow_session_metadata.c
```diff
@@ -6,26 +6,40 @@
...
+					size_t* cbor_data_len);
 
 int spotflow_session_metadata_send(void)
 {
 	const uint8_t* build_id = NULL;
 	uint16_t build_id_len = 0;
 
+	/* Generate device run ID once per boot */
+	if (device_run_id == 0) {
+		uint32_t rand_high = sys_rand32_get();
```
noorhaq, line 37: Shouldn't we keep it consistent with the CONFIG_SPOTFLOW_DEVICE_ID
Like we have in zephyr/src/net/spotflow_device_id.c

zephyr/samples/metrics/src/main.c
jmasek: no, it is intentional - user cannot use a custom frame, because the cloud supports only certain time frames. This is an intentional design decision.

zephyr/src/net/spotflow_session_metadata.c
jmasek: what do you mean? this is run id, not device id - it is a unique identifier of a run. it is different property than device id

.gitignore
jmasek: lgtm, removed.

zephyr/src/metrics/Kconfig
jmasek: good point.  Added `Not the overall log levels.` to Not the overall log levels.
While expecting that we will address the desing of spotflow internal logging in the next task, right?

zephyr/src/metrics/spotflow_metrics_cbor.c
jmasek: These are different fields - uptime is metadata and uptime is also value so it should be there in both fields.

zephyr/src/metrics/spotflow_metrics_cbor.c
jmasek: Interesting idea. Adding to the todo https://www.notion.so/spotflow/Device-SDK-suggestions-comments-etc-277400d453c580159bc7e32301a0c6be

zephyr/src/metrics/spotflow_metrics_backend.h
jmasek: What do you mean by mapping can cause issues? like converting uint8_t to int64_t?
I agree that it is a waste for call - on stack we allocate 8 bytes versus 1/2/4. However, if we are not going to expose aggregator and its internals to the user, the waste ends there - metric gets aggregated. 
btw. zxbor_encode exposes only functions for 32 and 64 int variants, so uint8 and uint16 would have to be converted anyway.

zephyr/src/metrics/spotflow_metrics_aggregator.c
jmasek: lgtm

zephyr/src/metrics/system/spotflow_metrics_system.h
jmasek: no we cannot, it is the limitation of the backend architecture. Basically, to be able to provide dashboards and reasonable query times, aggregation intervals must be well known.

zephyr/src/metrics/system/spotflow_reset_helper.c
jmasek: Good point, forwarded to the backend team.

zephyr/src/metrics/system/spotflow_reset_helper.c
jmasek: what dou you mean? reset cause is int, and each bit in this int is a different cause. Therefore, we need to check all possible causes if they were flagged.

zephyr/src/metrics/system/Kconfig
jmasek: These are different signals. I do not think that when sending a coredump (tens of kBs), adding a single message with a reset cause is a significant overhead.

noorhaq: True.

zephyr/src/metrics/spotflow_metrics_backend.h
noorhaq: Nope it is 2 separate parts.
One overflow issue as we don't have any for unsigned int like uint64 can cause issues.


Second when we will be saving them in future we should have some for smaller types to just to save space. 

Like the rest of the workflow will be the same just the log file and memory occupied will be low like say some ADC value or pwm value is between 0 to 255.

zephyr/samples/metrics/src/main.c
jmasek: yes, fixed

zephyr/src/metrics/spotflow_metrics_aggregator.c
jmasek: yes

zephyr/samples/metrics/src/main.c
jmasek: removed

zephyr/samples/metrics/src/main.c
jmasek: changed

zephyr/samples/metrics/README.rst
jmasek: legit, removed

CODING_STYLE.md
jmasek: good point, most likely not...

jmasek: removed

zephyr/src/net/spotflow_session_metadata.c
jmasek: why? it is static and used only in this file

zephyr/src/metrics/system/spotflow_reset_helper.c
jmasek: discussed - it is possible optimization in the future, but we won't do it now in the initial release. Adding to TODO/tech debt.

zephyr/src/metrics/spotflow_metrics_backend.h
jmasek: ok, regarding uint64_t - it is an issue I am aware of, and I will add it in a separate PR.

regarding smaller type - lets discuss it in chat

zephyr/src/metrics/system/spotflow_reset_helper.c
```diff
@@ -0,0 +1,107 @@
...
+		if (cause & reset_cause_map[i].flag) {
```
noorhaq, line 89: Can we use something like  `i = __builtin_ctz(cause);`
So it passes over the unused bits like say if only one flag bit is set we don't check all of them.
As at a time reset will be caused by just one reason not multiple. And once it finds it, it breaks the loop.

zephyr/src/net/spotflow_session_metadata.c
roberthusak: The new `CODING_STYLE.md` says that all global static variables should have the prefix `g_`. Or was it meant for non-static global variables?

zephyr/samples/metrics/credentials-sample.conf
jmasek: updated

zephyr/samples/metrics/README.rst
jmasek: cleaned up, not adding sample - not sure about it.

.editorconfig-checker.json
jmasek: I do not see any added calue if editorconfig will be checking readme file

zephyr/src/net/spotflow_session_metadata.c
jmasek: adjusted coding_style. Incosistency in g_ prefix will be fixed in next commits

zephyr/src/metrics/system/spotflow_metrics_system_heap.c
jmasek: keeping it that way until support for uint64 is added

zephyr/src/net/spotflow_session_metadata.c
jmasek: changed to `%"PRIu64`

zephyr/src/metrics/system/spotflow_metrics_system_stack.c
jmasek: removed

zephyr/samples/metrics/src/main.c
jmasek: legit. Will do in separate PR.

zephyr/src/coredumps/spotflow_coredumps_net.c
jmasek: actually, the call will be back in the main loop where the mqtt message will be prepared again so I think there will be few cycles. If you do not mind, keeping it that way right now.

zephyr/src/metrics/spotflow_metrics_backend.h
jmasek: discussed in chat. closing

zephyr/samples/metrics/credentials-sample.conf
jmasek: added

zephyr/src/metrics/spotflow_metrics_aggregator.c
jmasek: added on todo to allow disable jitter

zephyr/src/metrics/spotflow_metrics_types.h
jmasek: yes

zephyr/src/metrics/spotflow_metrics_types.h
jmasek: no removed

zephyr/src/metrics/system/spotflow_reset_helper.c
noorhaq: @jmasek Just a suggestion here we can skip it. Or add it in optimisation in future because if we remove the reset string from device then it won't be required as it will become a cloud side dependency

zephyr/src/logging/spotflow_log_net.h
noorhaq: @jmasek Just FYI. Because i think it was missed.

zephyr/src/metrics/system/spotflow_reset_helper.c
jmasek: I am not sure if it is reasonable - this loop does 16 iterations. If I were to use __builtin_ctz, the code will be much more complex, less readable, and we will savea  few iterations. It does not feel like a good deal.

zephyr/src/metrics/spotflow_metrics_cbor.c
jmasek: good point, renamed

jmasek: good point, renamed

zephyr/src/metrics/spotflow_metrics_cbor.c
jmasek: no, it uses different type/struct

zephyr/src/metrics/spotflow_metrics_registry.c
jmasek: ofc! removed

zephyr/src/metrics/spotflow_metrics_aggregator.c
jmasek: one was removed

zephyr/src/metrics/spotflow_metrics_backend.h
jmasek: reworked return codes, please check them

zephyr/src/metrics/spotflow_metrics_backend.h
jmasek: updated

zephyr/src/metrics/spotflow_metrics_net.h
jmasek: update, btw. unification - as proposed by Noor is up for discussion

zephyr/src/metrics/spotflow_metrics_backend.c
jmasek: moved

CODING_STYLE.md
jmasek: added

zephyr/src/logging/spotflow_log_net.h
jmasek: you meant I missed spotflow prefix, right?

zephyr/src/metrics/spotflow_metrics_cbor.c
jmasek: fixed

zephyr/src/metrics/spotflow_metrics_aggregator.c
jmasek: good point, fixed

CODING_STYLE.md
roberthusak: It's still present in all the newly added files.

zephyr/src/metrics/spotflow_metrics_net.h
roberthusak: Btw. my original comment was aimed also at the first sentence (`Memory Ownership: Processor thread...`) because it's AFAIK not true (according to the function implementation).

zephyr/src/metrics/system/spotflow_reset_helper.c
roberthusak: Btw. isn't `__builtin_ctz` specific to GCC? AFAIK Zephyr aims to have support for other compilers (including proprietary ones) as well and using GCC-specific features might break the support for them.

zephyr/src/metrics/spotflow_metrics_net.h
jmasek: I see, removed

CODING_STYLE.md
jmasek: omg, I removed only from coding style, not from files. removed

roberthusak (dismissed):

roberthusak (approved):

PR #55, opened 2026-01-08
Add list of supported boards for documentation and CI

roberthusak:


zephyr/ci/spotflowup.ps1
```diff
@@ -0,0 +1,1163 @@
...
+                break
+            }
+        }
+        catch {
+            continue
+        }
+    }
+
+    if (-not $pythonCmd) {
+        Exit-WithError "Python not found" $details `
```
jmasek: does $details has any sense here, will it be defined?

.github/workflows/quickstart.yml
```diff
@@ -0,0 +1,303 @@
...
+      # Windows runner contains Python 3.9, update to the minimum required version
+      - name: Install Python
+        if: runner.os == 'Windows'
+        uses: actions/setup-python@v6
+        with:
+          python-version: '3.10'
+
+      # Needed for Zephyr SDK installation, not available on Windows runner
+      - name: Install wget
+        if: runner.os == 'Windows'
```
noorhaq, line 124: Why are we installing on all OSes? Like windows, macOS and ubuntu. Wouldn't it be easier to maintain just on one? Like Ubuntu.

zephyr/ci/boards.yml
```diff
@@ -0,0 +1,394 @@
...
+      - id: esp32c3_devkitc
+        name: ESP32-C3-DevKitC
+        sdk_toolchain: riscv64-zephyr-elf
+
+      - id: esp32c3_devkitm
+        name: ESP32-C3-DevKitM
+        sdk_toolchain: riscv64-zephyr-elf
+        build_samples: ["logs"]
+
+      - id: esp32c3_rust
```
noorhaq, line 34: Is rust tool chain required? Or just for future use.

zephyr/ci/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Workspace Setup Scripts
+
+`spotflowup.sh` and `spotflowup.ps1` automate the west workspace setup so that the user can easily compile our samples for the selected board.
+
+Apart from differences in certain Bash and PowerShell commands, the scripts have the same functionality.
+Operating systems supported:
+
+- `spotflowup.sh` - Linux and macOS
+- `spotflowup.ps1` - Linux, macOS, and Windows
```
noorhaq, line 51: Do we need to support windows as zephyr itself uses the Windows subsystem for linux for zephyr support?

zephyr/samples/coredumps/prj.conf
```diff
@@ -1,32 +1,33 @@
-# NXP FRDM-RW612 board uses Wi-Fi by default. If you want to use Ethernet instead,
-# please use the following options:
+# Boards that support both Wi-Fi and Ethernet use Wi-Fi by default.
+# If you want to use Ethernet instead, please use the following option:
 # CONFIG_SPOTFLOW_USE_ETH=y
```
noorhaq, line 3: We have to add some check so only ETH or wifi is enabled or some warning is displayed as on some stacks/boards if both are enabled at the same time required buffers for the net stack increases and we can hit on board memory limits.

zephyr/samples/coredumps/prj.conf
```diff
@@ -1,32 +1,33 @@
...
 
-#configure what should be included in the coredump
-#CONFIG_DEBUG_COREDUMP_MEMORY_DUMP_MIN=y
-CONFIG_DEBUG_COREDUMP_THREADS_METADATA=y
-CONFIG_DEBUG_COREDUMP_MEMORY_DUMP_THREADS=y
-
-#configure spotflow internal logging levels
-CONFIG_SPOTFLOW_MODULE_DEFAULT_LOG_LEVEL=4
+# Enable logging
+CONFIG_PRINTK=y
```
noorhaq: Won't it better to use `CONFIG_LOG_PRINTK` instead of `CONFIG_PRINTK` so the logging subsystem i.e. the messages pass through the spotflow component?

.github/workflows/quickstart.yml
roberthusak: It would be easier. However, if we tested the scripts only on one OS, we wouldn't be sure if they work on other ones. There might subtle differences in file paths etc.

zephyr/ci/boards.yml
roberthusak: Rust toolchain isn't needed. AFAIK the board was designed so that it's easy to write Rust projects for it. However, you can write firmware for it normally in C.

zephyr/ci/README.md
roberthusak: I was considering it too when I was starting with Zephyr development. However, they have this big disclaimer in [Getting Started](https://docs.zephyrproject.org/latest/develop/getting_started/index.html): 😅

> Due to issues finding executables, the Zephyr Project doesn’t currently support application flashing using the [Windows Subsystem for Linux (WSL)](https://msdn.microsoft.com/en-us/commandline/wsl/install_guide) (WSL).
>
> Therefore, we don’t recommend using WSL when getting started.

So it's questionable how popular WSL for Zephyr development actually is. It seems that the primary way of Zephyr development on Windows (according to Getting Started) is without WSL. Furthermore, nRF Connect SDK VS Code extension uses this approach, too.

zephyr/samples/coredumps/prj.conf
roberthusak: I think this should be already handled by the Kconfig of each sample, see https://github.com/spotflow-io/device-sdk/blob/feature/board-list/zephyr/samples/logs/Kconfig#L22-L43

Btw. we only have two boards in the list that support both Wi-Fi and ETH (ESP32-Ethernet-Kit and FRDM-RW612) and the logging sample compiles for both of them.

jmasek (changes requested):

zephyr/samples/coredumps/prj.conf
roberthusak: Good point 👍 Turns out that we can leave it out altogether, because `CONFIG_PRINTK` is selected by `CONFIG_LOG` and `CONFIG_LOG_PRINTK` is enabled by default when `CONFIG_PRINTK` is enabled.

zephyr/ci/spotflowup.ps1
roberthusak: It doesn't. I removed it, thanks.

zephyr/ci/README.md
jmasek: I personally work on Windows and I work with standard powershell - no wsl without any issues. Therefore, I do not see reason for using wsl

jmasek (approved):

PR #56, opened 2026-01-18
ESP IDF component testing.

noorhaq:
Added tests for esp idf components.


esp_idf/spotflow/device_sdk/examples/unit_test/test/main/example_unit_test.c
```diff
@@ -0,0 +1,40 @@
+/* Example test application for testable component.
+
+   This example code is in the Public Domain (or CC0 licensed, at your option.)
+
+   Unless required by applicable law or agreed to in writing, this
+   software is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+   CONDITIONS OF ANY KIND, either express or implied.
+*/
```
jmasek: different license than the rest of the repo

esp_idf/spotflow/device_sdk/test/cbor/test_cbor.c
```diff
@@ -0,0 +1,117 @@
...
+
+TEST_CASE("CBOR encodes a simple log correctly", "[spotflow][cbor]")
+{
+    const char* template = "Test log template";
+    char body[] = "Hello CBOR";
+    struct message_metadata meta = {
+        .sequence_number = 1,
+        .uptime_ms = 1000,
+        .source = "unit_test"
+    };
```
jmasek: 1. why is severity in metadata it seems that in cbor serialization, the parameter from spotflow_log_cbor_send is taken and metadata's severity is ignored. 
2. Could you please also test that severity is serialized correctly?

esp_idf/spotflow/device_sdk/test/example/unit_test/README.md
```diff
@@ -0,0 +1,24 @@
+# Test skeleton
```
jmasek, line 1: Isn't it confusing to have unit_test in examples? Maybe we could have it next to the examples folder, what do you think?

esp_idf/spotflow/device_sdk/examples/unit_test/test/main/example_unit_test.c
mashtak: different license than repo one, please remove

esp_idf/spotflow/device_sdk/test/log_backend/test_log_backend.c
```diff
@@ -0,0 +1,169 @@
...
+
+    const char* log_msg = "Test severity log";
+
+    // Test for ERROR severity
+    ESP_LOGE("TAG", "%s", log_msg);
+    read_queue_and_verify(log_msg);  // First verify the message is logged
+
+    // Now check if the CBOR data has the correct key-value pairs
+    uint8_t* cbor_data = (uint8_t*)queue_msg.ptr;  // The CBOR encoded message
+    size_t cbor_len = queue_msg.len;
```
mashtak: read_queue_and_verify calls spotflow_queue_free(&queue_msg); that frees queue_msg.ptr but then you do uint8_t* cbor_data = (uint8_t*)queue_msg.ptr;  and use cbor_data

esp_idf/spotflow/device_sdk/test/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
+idf_component_register(
+    SRC_DIRS
+        "queue"
+        "cbor"
+        # "log_backend"
```
mashtak: why log_backend test is commented out?

esp_idf/spotflow/device_sdk/test/log_backend/test_log_backend.c
```diff
@@ -0,0 +1,169 @@
...
+    read_queue_and_verify(large_msg);
+
+    free(large_msg);  // Free the allocated memory
+
+    // Cleanup
+    esp_log_set_vprintf(NULL);  // Reset to default vprintf behavior
+}
+
+// Helper function to check if CBOR data contains a specific key-value pair
+static bool contains_cbor_key(uint8_t* cbor_data, size_t cbor_len, uint32_t key, uint32_t expected_value) {
```
mashtak: What if some part of value will contain key? then you will treat value as key; therefore, creating a false negative. Is there a reason you are using byte-by-byte comparison instad of cbor parser?

esp_idf/spotflow/device_sdk/test/cbor/test_cbor.c
```diff
@@ -0,0 +1,117 @@
...
+    const char* template = "Template";
+    char body[] = "Msg";
+    struct message_metadata meta = { .sequence_number = 1, .uptime_ms = 100, .source = "" };
+
+    cbor_buf = spotflow_log_cbor(template, body, LOG_SEVERITY_WARN, &cbor_len, &meta);
+
+    TEST_SPOTFLOW_ASSERT_TRUE(cbor_buf != NULL);
+
+    // The CBOR should encode empty string for "source"
+    TEST_SPOTFLOW_ASSERT_TRUE(memmem(cbor_buf, cbor_len, "", 0) != NULL);
```
mashtak: does this make any sense? basically searching for zero-legth string in buffer - always return pointer to buffer

esp_idf/spotflow/device_sdk/test/cbor/test_cbor.c
```diff
@@ -0,0 +1,117 @@
...
+    cbor_buf = spotflow_log_cbor(template, body, LOG_SEVERITY_INFO, &cbor_len, &meta);
+
+    TEST_SPOTFLOW_ASSERT_TRUE(cbor_buf != NULL);
+    TEST_SPOTFLOW_ASSERT_LESS_OR_EQUAL(MAX_CBOR_LEN, cbor_len);
+
+    // Basic content check: body string is present
+    TEST_SPOTFLOW_ASSERT_TRUE(memmem(cbor_buf, cbor_len, "Hello CBOR", strlen("Hello CBOR")) != NULL);
+
+    // Check metadata sequence number (encoded as integer somewhere in CBOR)
+    TEST_SPOTFLOW_ASSERT_TRUE(memmem(cbor_buf, cbor_len, &meta.sequence_number, 1) != NULL);
```
mashtak: why you are using size 1 and not the size of sequence_number

esp_idf/spotflow/device_sdk/test/port/test_wrapper.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
```
mashtak: is there a reason for using #pragma once instead of standard #ifndef used in normal code?

esp_idf/spotflow/device_sdk/LICENSE.md
```diff
@@ -0,0 +1 @@
+../../../LICENSE.MD
```
mashtak, line 1: inconsistent file naming - this file is .md, referenced is .MD

mashtak (changes requested):

jmasek (changes requested):

jmasek:
Generally, I would prefer to test the CBOR using a deserializer - either the same one used for serialization or a different one (preferred if possible). Searching bytes in serialized cbor can lead to fragile tests. e.g., if you serialize uint64_t to CBOR and its value is smaller than 255, it will get encoded into a single byte - optimization done by cbor, so you won't be able to do the mem search.

esp_idf/spotflow/device_sdk/test/CMakeLists.txt
noorhaq: Because it has some dependency on cloud testing visibility due to which it is commented out for now.

noorhaq:
Yeah I will change this.

esp_idf/spotflow/device_sdk/test/log_backend/test_log_backend.c
noorhaq: Added cbor parser.

esp_idf/spotflow/device_sdk/test/cbor/test_cbor.c
noorhaq: Fixed.

esp_idf/spotflow/device_sdk/test/cbor/test_cbor.c
noorhaq: Fixed.

esp_idf/spotflow/device_sdk/LICENSE.md
noorhaq: This is like for reference due to some reason the esp idf component website wasn't picking up the license.MD so used this one and it worked so didn't touch it afterwards same is for changelog as well. This was just a symbolink link created for the UI.

esp_idf/spotflow/device_sdk/test/cbor/test_cbor.c
noorhaq: Okay this one went through the cracks. I was looking into the backend logic as well and their I missed it i kept on using the severity as a seperate argument instead of using the metadata. I will fix it.

esp_idf/spotflow/device_sdk/test/example/unit_test/README.md
noorhaq: This is like a test skeleton like the actual tests are in the test folder this example just shows how to use them or for us testing. It's a copy paste of the esp idf testing example. 
With changes to cmake file

So, we can move it or remove it entirely from the component and just use it internally.

esp_idf/spotflow/device_sdk/test/cbor/test_cbor.c
noorhaq: Fixed.

esp_idf/spotflow/device_sdk/test/example/unit_test/README.md
jmasek: and why not move it to test/example? I understanding of example folder are examples for users - how to user our sdk. Tests are something that is rather internal so I would prefer to keep it somewhere else than the main entry point for users

jmasek (commented):
unfortunately, the pipeline is failing, can you fix it before merging?

noorhaq:
Pipeline is fixed now

esp_idf/spotflow/device_sdk/test/example/unit_test/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+This project is only for testing the componnent that's why it's main is inside the `test/example_unit_test.c`. For testing we are using the esp idf unity framework.
```
roberthusak:
```suggestion
This project is only for testing the component that's why it's main is inside the `test/main/example_unit_test.c`. For testing we are using the ESP-IDF Unity framework.
```

esp_idf/spotflow/device_sdk/test/log_backend/test_log_backend.c
```diff
@@ -0,0 +1,253 @@
...
+	read_queue_and_verify(large_msg);
+
+	free(large_msg); // Free the allocated memory
+
+	// Cleanup
+	esp_log_set_vprintf(NULL); // Reset to default vprintf behavior
+}
+
+// Helper function to check if CBOR data contains a specific key-value pair
+static bool contains_cbor_key(const uint8_t* cbor_data, size_t cbor_len, uint32_t key,
```
roberthusak: Shouldn't this function be unified with `contains_cbor_uint_key` from `test_cbor.c`?

esp_idf/spotflow/device_sdk/test/cbor/test_cbor.c
```diff
@@ -0,0 +1,289 @@
...
+	cbor_len = 0;
+}
+
+void tearDown(void)
+{
+	if (cbor_buf) {
+		free(cbor_buf);
+		cbor_buf = NULL;
+	}
+}
```
roberthusak, line 40: Neither `setUp` nor `tearDown` is used in the file.

esp_idf/spotflow/device_sdk/test/cbor/test_cbor.c
```diff
@@ -0,0 +1,289 @@
...
+				if (cbor_value_calculate_string_length(&element, &value_len) ==
+				    CborNoError) {
+					if (value_len == expected_len) {
+						if (value_len == 0U) {
+							match = true;
+						} else {
+							char* buf = (char*)malloc(value_len + 1U);
+							if (buf != NULL) {
+								size_t copied = value_len + 1U;
+								if ((cbor_value_copy_text_string(
```
roberthusak: I think that 8 levels of nesting scream for a reorganization or a helper function. Wouldn't it be useful to have a helper function that will look up the value in a map? This function could be then used both in `contains_cbor_text_value` and `contains_cbor_uint_key` (which I'd recommend to rename to `contains_cbor_uint_value` - both are looked up for by a key).

roberthusak (commented):
I added some minor comments, seems good to me otherwise 👍

I couldn't find running tests in CI, are you planning to add it? I suppose it should work in QEMU.

esp_idf/spotflow/device_sdk/test/cbor/test_cbor.c
noorhaq: Nope these are called internally by esp idf framework like to cleanup and setup tests. Like a constructor and destructor.

esp_idf/spotflow/device_sdk/test/log_backend/test_log_backend.c
noorhaq: Nope this one is testing another aspect of it. Like the test cbor is handling everything from the struct formation and onward. This one is doing a sanity check from the basic log formation
i.e. the workflow if I connect is log message -> struct division like severity, time etc stuff -> convertion to cbor and return
The log test do it on the overall message like error message and check if the formation was correct or not. Not complete in depth testing just a slight check on all params.

esp_idf/spotflow/device_sdk/test/cbor/test_cbor.c
noorhaq: Done. On the nesting this a helper function would not fix. Like these are checks for strings like I didn't add them under a single if block because this way is helpful in debugging like you can jump each step to see what might be causing it to fail. And all are cbor library internal functions so we won't be able to use them.
I have renamed the function to uint_value

noorhaq:
> I added some minor comments, seems good to me otherwise 👍
> 
> I couldn't find running tests in CI, are you planning to add it? I suppose it should work in QEMU.

Yep it does work in qemu like locally we can test it in qemu as well. And I will have a short call with you regarding it's addition to the jobs as well. Like we can run all tests or some using the interface that opens up. But it is interactive so we can trigger it respectively.

esp_idf/spotflow/device_sdk/test/log_backend/test_log_backend.c
```diff
@@ -0,0 +1,253 @@
...
+	setUP_init();
+
+	const char* log_msg = "Test severity log";
+
+	// Test for ERROR severity
+	ESP_LOGE("TAG", "%s", log_msg);
+	read_queue_and_verify(log_msg); // First verify the message is logged
+
+	// Now check if the CBOR data has the correct key-value pairs
+	uint8_t* cbor_data = (uint8_t*)queue_msg.ptr; // The CBOR encoded message
```
jmasek: read_queue_and_verify frees the queue_msg but you use it there, so you are reading freed variable, or am I missing something.

esp_idf/spotflow/device_sdk/test/CMakeLists.txt
jmasek: can you please add explanation why it is commented out?

esp_idf/spotflow/device_sdk/test/example/unit_test/README.md
```diff
@@ -0,0 +1,24 @@
+# Test skeleton
+
+This project is only for testing the component that's why it's main is inside the `test/main/example_unit_test.c`. For testing we are using the ESP-IDF Unity framework.
+## Configurations
+
+## Instructions for running the example.
+- In your terminal from inside this folder run
+```bash
+cd test/
```
jmasek, line 9: why the structure is that deep? I see test/example/unit_test/test? is there a reason to have an empty parent folder /unit_test?

jmasek (commented):
Last few small comments. When fixed and discussion about CI is finished you are good to go.

esp_idf/spotflow/device_sdk/test/example/unit_test/README.md
noorhaq: Yes because this will be like the normal like how the final project will look like.
Like the hierchy will be similar only it won't be named as unit test it will be some project name.
With a test folder inside to runt he tests. 

Unit_test can be replaced with a python script as well. Depending on the use case.
https://github.com/espressif/esp-idf/tree/master/examples/system/unit_test

esp_idf/spotflow/device_sdk/test/log_backend/test_log_backend.c
noorhaq: Nope this whole thing was commented out in Cmake because it is like a unfinished test case. 
Like this was supposed to check the return from the cloud like to make sure no issue arrived in the mqtt buffer or any networking buffer. So those calls are missing in between.
I didn't remove it just kept it unfinished to integrate with cloud testing when we do it in future.
The Cbor test is already doing all the verification that we are doing here. Like it will look duplicate tests.

roberthusak (dismissed):

esp_idf/spotflow/device_sdk/test/log_backend/test_log_backend.c
jmasek: Then it should be clearly noted that it is unfinished, and also the code logic is broken. Also, I do not like keeping invalid code - even if not used.

esp_idf/spotflow/device_sdk/test/CMakeLists.txt
jmasek: still missing an explanation in the file

jmasek: You removed it instead of adding a comment?

noorhaq: [Log_backend.c](https://github.com/noorhaq/device-sdk/blob/239fe65cbc605b26cf89ba7c45fe0c55f83f5c3f/esp_idf/spotflow/device_sdk/test/log_backend/test_log_backend.c#L2)
After discussion with Robert added the comment in file and removed it from Cmake instead of commenting it out.
It needs a mock mqtt server or some other way to view the cloud received messaged. Otherwise it is similar to test_cbor.
It is testing the overall workflow.
-> Device sends cbor message to cloud.
-> Cloud receives it. Or mock server receives it. // Missing this part currently so this is in Todo. Inside the file,
-> Check the msg if it is correct or not, by reusing the queue struct or we can create a new one. Like depending on future.
It checks the mqtt buffers and other net buffers to see if nothing was truncated during transport.

esp_idf/spotflow/device_sdk/test/log_backend/test_log_backend.c
noorhaq: It does now inside the file we have added comment at the top of the file. We can remove the whole file as it is not being compiled currently.

roberthusak: @jmasek We discussed this with Noor and ended up with a "TODO" comment and "unfinished" comment at the top of the file, but you're right that it's probably unsufficient - no comment is good enough to justify dead code.

@noorhaq I still see the value of this test file beyond checking serialization - it checks that the logging backend pipeline works - e.g., when I call `ESP_LOGE(...)`, the test checks that an error log is emitted. Therefore, I think that the ideal solution would be to make the code work even without the MQTT connection. That is, for now, just check that the right messages are delivered to the queue. If that's not possible (or it would be excessively complicated), remove the file.

Regarding MQTT, we might add end-to-end tests later. This PR should be AFAIK focused on unit tests anyway.

noorhaq: Done removed the file.
Well we are already checking that in test cbor like this will only be helping to create the struct metadata and add severity their except that tests would exactly be the same. Specially the cbor. 
Like it will be to verify that the metadata struct is crafted correctly. 

So, without mqtt in anyway i don't see any value in this file or test process as it would be very superficial and repetitive.

jmasek (approved):

PR #58, opened 2026-01-21
added resolve version action

jmasek:


noorhaq (approved):

.github/workflows/esp_idf_component_registry.yml
```diff
@@ -5,6 +5,10 @@ on:
       - 'v*'            # Trigger on tags starting with v (e.g., v1.0.0)
     paths:
       - 'esp_idf/**'    # Trigger on changes within the esp_idf directory
+  pull_request:
+    paths:
+      - 'esp_idf/**'    # Trigger on PRs with changes in esp_idf directory
+      - '.github/workflows/**'  # Trigger on PRs with workflow changes
```
noorhaq: After testing we should remove it. So in case no updates are added and other versions are released say for zephyr or stm we don't trigger it

noorhaq (dismissed):

.github/workflows/esp_idf_component_registry.yml
jmasek: removed

PR #66, opened 2026-02-25
Over-the-air (OTA) updates

roberthusak:


zephyr/samples/ota/prj.conf
```diff
@@ -0,0 +1,55 @@
...
+CONFIG_NET_UDP=y
+
+# Enable a persistent settings backend so OTA worker results can be saved and reported.
+CONFIG_NVS=y
+CONFIG_SETTINGS_NVS=y
+
+# Increase main stack size for the sample
+CONFIG_MAIN_STACK_SIZE=4096
+
+# TODO: Fixes stack overflow (seems to happen when NVS is enabled), propagate to Kconfig
```
roberthusak: I'll fix that in a separate PR (doesn't happen only with OTA).

zephyr/Kconfig
```diff
@@ -62,7 +64,10 @@ config SPOTFLOW_GENERATE_BUILD_ID
 		Build ID allows to match core dumps with the original .elf file.
 
 configdefault MBEDTLS_SSL_MAX_CONTENT_LEN
-	default 4096 if SPOTFLOW
+	# TODO: Remove if we manage to support max_fragment_length in the CDN
```
roberthusak: I'll discuss this with Tom.

roberthusak: Won't do - `max_fragment_length` isn't supported by Envoy. Will be solved by TLS 1.3 in the future.

Alternatively, we might enforce the usage of `Range` headers of artifact downloads if `MBEDTLS_SSL_MAX_CONTENT_LEN` is lower than 16 KB. That would be less efficient and more brittle though.

zephyr/src/ota/spotflow_ota.c
```diff
@@ -0,0 +1,443 @@
...
+	spotflow_ota_state_reset();
+	spotflow_ota_worker_reset();
+#if IS_ENABLED(CONFIG_SPOTFLOW_OTA_AUTO_HANDLE_MAIN_FIRMWARE)
+	spotflow_ota_fw_main_reset();
+#endif
+}
+
+static void handle_ota_c2d_msg(uint8_t* payload, size_t len)
+{
+	struct spotflow_ota_cbor_c2d_msg msg;
```
jmasek: Problem: struct spotflow_ota_cbor_c2d_msg is allocated on the MQTT processing thread stack. At the permitted maximum of 32 artifacts, it exceeds 8 KB, while the processing thread defaults to 4 KB.
Why it matters: Receiving a valid maximum-size manifest can overflow the thread stack and crash the device. Repeated cloud delivery could create a reboot loop.
Suggested fix: Move decode storage to bounded static or dedicated OTA storage, or provide a correctly sized dedicated thread. Add maximum-artifact stack-usage validation on hardware.

zephyr/src/ota/firmware/spotflow_ota_fw_main.c
```diff
@@ -0,0 +1,816 @@
...
+	}
+
+	rc = spotflow_ota_state_set_main_firmware_result(SPOTFLOW_OTA_RESULT_SUCCEEDED, &state);
+	if (rc < 0) {
+		return rc;
+	}
+
+	notify_main_firmware_state(&state);
+
+	rc = spotflow_ota_persistence_clear_probation();
```
jmasek: Shouldn't be probation cleared after the persistence - also in case of rollback? 
A Settings failure or power loss between those operations leaves a pending persisted attempt with no probation record. The confirmed or rolled-back update can no longer be reconciled -> A stale probation is recoverable because startup can recognize the terminal result and clear it.

zephyr/src/ota/platform/spotflow_ota_platform.c
```diff
@@ -0,0 +1,114 @@
...
+#include <zephyr/storage/flash_map.h>
+#include <zephyr/sys/reboot.h>
+
+#include "ota/platform/spotflow_ota_platform.h"
+
+static struct flash_img_context upload_ctx;
+
+int spotflow_ota_platform_request_test_upgrade(void)
+{
+	return boot_request_upgrade(BOOT_UPGRADE_TEST);
```
jmasek, line 20: SPOTFLOW_OTA_AUTO_HANDLE_MAIN_FIRMWARE only depends on BOOTLOADER_MCUBOOT. It does not guarantee rollback support. you assume that an unconfirmed image will be reverted. That assumption is false in non-revertible MCUboot modes.

zephyr/src/ota/core/spotflow_ota_state.c
```diff
@@ -0,0 +1,809 @@
...
+			current_attempt.attempt_id = probation->attempt_id;
+
+			if (has_attempt && attempt != NULL &&
+			    attempt->attempt_id == probation->attempt_id) {
+				current_attempt.update.artifact_count = attempt->artifact_count;
+				memcpy(current_attempt.results, attempt->artifact_results,
+				       sizeof(current_attempt.results));
+			} else {
+				current_attempt.update.artifact_count =
+					probation->artifact_index + 1;
```
jmasek: shouldn't this be validated against CONFIG_SPOTFLOW_OTA_MAX_ARTIFACTS?

jmasek: maybe also in validate_probation()

jmasek: in case the settings returns corrupted record

zephyr/src/ota/firmware/spotflow_ota_fw_main.c
```diff
@@ -0,0 +1,816 @@
...
+	notify_main_firmware_phase(SPOTFLOW_OTA_PHASE_PENDING_REBOOT);
+
+	rc = spotflow_ota_state_finish_main_firmware_prereboot(NULL);
+	if (rc < 0) {
+		LOG_ERR("Failed to finalize main firmware before reboot: %d", rc);
+		return fail_main_firmware();
+	}
+
+	spotflow_ota_platform_reboot();
+
```
jmasek: if the abort arrives after check, success is returned but the device will reboot and boot into the new firware anyway. This is a race between the OTA worker and spotflow_abort_main_firmware_update(). Is it ok?

zephyr/src/ota/downloader/spotflow_ota_downloader_transport_socket.c
```diff
@@ -0,0 +1,252 @@
...
+	if (ctx->callback_err != 0) {
+		return ctx->callback_err;
+	}
+
+	if (downloader_is_canceled(ctx->downloader)) {
+		ctx->callback_err = -ECANCELED;
+		return -ECANCELED;
+	}
+
+	if (rsp->http_status_code != 0) {
```
jmasek, line 225: Shouldn't you also check 
rsp->cr_present
Content-Range start equals the requested range_start
Range end is valid
Total artifact size is consistent across retries

If a CDN, proxy, or server returns a 206 response for the wrong byte range, the downloader appends those bytes at the current stream position.
This can duplicate or omit data.

zephyr/src/ota/Kconfig
```diff
@@ -0,0 +1,77 @@
+menuconfig SPOTFLOW_OTA
+	bool "Enable Spotflow over-the-air (OTA) updates"
+	depends on SPOTFLOW
+	depends on SPOTFLOW_TRANSPORT_MQTT
+	select SETTINGS
```
jmasek: this does not require a persistent or runtime Settings backend.
When Settings is enabled without NVS, ZMS, FCB, file, retention, custom, or runtime storage, Zephyr can select SETTINGS_NONE.

In that configuration:
- settings_subsys_init() returns success.
- OTA initialization reports success.
- Every later settings_save_one() returns -ENOENT.

zephyr/src/ota/core/spotflow_ota_state.c
```diff
@@ -0,0 +1,809 @@
+#include <errno.h>
```
jmasek, line 7: missing copyright header

jmasek:
please add the changelog and make sure all the new files conform CODING_STYLE.md
mainly 
- New OTA C and header files omit the required copyright header.
- Source files generally do not include their own header first.

zephyr/include/spotflow/downloader.h
```diff
@@ -0,0 +1,151 @@
...
+	struct k_mutex mutex;
+	bool cancel_requested;
+	struct k_sem resume_sem;
+};
+
+/** Define and statically initialize a @ref spotflow_downloader. */
+#define SPOTFLOW_DEFINE_DOWNLOADER(name)                                \
+	struct spotflow_downloader name = {                             \
+		.state = SPOTFLOW_DOWNLOADER_STATE_INACTIVE,            \
+		.mutex = Z_MUTEX_INITIALIZER(name.mutex),               \
```
jmasek: is really needed to use macros that are * @cond INTERNAL_HIDDEN `Z_MUTEX_INITIALIZER and Z_SEM_INITIALIZER.`?

jmasek:
do you consider adding automatic pipeline build of ota sample for some boards?

zephyr/src/ota/core/spotflow_ota_state.c
```diff
@@ -0,0 +1,809 @@
...
+
+void spotflow_ota_state_reset(void)
+{
+	k_mutex_lock(&state_mutex, K_FOREVER);
+	clear_attempt(&current_attempt);
+	clear_pending_attempt();
+	k_mutex_unlock(&state_mutex);
+}
+
+int spotflow_ota_state_init_from_persistence(const struct spotflow_ota_persisted_attempt* attempt,
```
jmasek, line 55: could you please add some diagram with these transitions - reboot -> probation -> next firmware -> reboot etc.

zephyr/src/ota/core/spotflow_ota_state.c
```diff
@@ -0,0 +1,809 @@
...
+	if (has_attempt && attempt != NULL && attempt->attempt_id != 0) {
+		current_attempt.active = true;
+		current_attempt.attempt_id = attempt->attempt_id;
+		current_attempt.update.artifact_count = attempt->artifact_count;
+		current_attempt.actionable_cancellation = attempt->actionable_cancellation;
+		current_attempt.has_attempt_error = attempt->has_attempt_error;
+		current_attempt.attempt_error = attempt->attempt_error;
+		memcpy(current_attempt.results, attempt->artifact_results,
+		       sizeof(current_attempt.results));
+		advance_current_artifact(&current_attempt);
```
jmasek: After the persistance, the artifacts are not restored, is it correct? Therefore, after reboot they remain zero-initialized.
The probation restoration separately populates:
current_attempt.main_firmware_artifact

The failure scenario is:
1. Manifest contains multiple artifacts.
2. Main firmware is not the last artifact.
3. Main firmware downloads and the device reboots.
4. Attempt results and probation are restored.
5. New image is confirmed.
6. Confirmation marks the main artifact successful.
7. A later artifact remains pending, so the worker is woken.
8. spotflow_ota_state_get_worker_job() reads the later artifact from update.artifacts[index].
9. copy_artifact_out() correctly copies that structure, but the structure is zero-filled.

A duplicate manifest currently does not repair this because same-ID messages are ignored at spotflow_ota_state.c:147-155.

zephyr/src/ota/spotflow_ota.c
roberthusak: Good point, I made it `static`.

zephyr/src/ota/firmware/spotflow_ota_fw_main.c
roberthusak: Thanks, I fixed this issue together with another persistence-related one.

zephyr/src/ota/platform/spotflow_ota_platform.c
roberthusak: Added enforcement of a rollback-capable MCUboot mode.

zephyr/src/ota/core/spotflow_ota_state.c
roberthusak: I added the validation to `validate_probation`. That should be sufficient, because `spotflow_ota_state` will work only with an already validated value.

zephyr/src/ota/firmware/spotflow_ota_fw_main.c
roberthusak: Good point, there were multiple issues with aborting and pausing as well. The logic is now refactored and better covered by tests.

zephyr/src/ota/core/spotflow_ota_state.c
```diff
@@ -0,0 +1,940 @@
...
+		current_attempt.running_artifact_index = index;
+		k_mutex_unlock(&state_mutex);
+		return true;
+	}
+
+	k_mutex_unlock(&state_mutex);
+	return false;
+}
+
+int spotflow_ota_state_apply_artifact_result(size_t artifact_index, enum spotflow_ota_result result,
```
jmasek: this function is not  not reached on every error path. -> possible of logical deadlock.  
The fix should separate handler execution from durable result persistence and guarantee that every claimed job eventually reaches one of:
- Retryable
- Terminal
- Explicitly abandoned with a terminal failure

Example: installed-version load failure
At spotflow_ota_worker.c:148-152:
int rc = load_artifact_result(job, &result);
if (rc < 0) {
	LOG_ERR(...);
	return rc;
}
load_artifact_result() first reads the installed version from Settings. If that fails, the function returns before applying any artifact result.
State remains:
artifact_running = true
result[index] = PENDING
The outer worker loop then runs again.
spotflow_ota_state_get_worker_job() refuses to return the artifact because artifact_running is true. The attempt is also not terminal because its result is still pending.
The worker therefore finds:
- No available job
- No terminal attempt to persist
- No pending attempt that can be promoted
It exits the inner loop and waits on its semaphore indefinitely.

Example: successful handler followed by persistence failure
The delegated handler can successfully update an external device and return SUCCEEDED.
The worker then tries to persist the installed version at spotflow_ota_worker.c:159-165. If that write fails, the worker returns before applying the successful result.
This is more complicated than simply retrying the entire job because rerunning the handler may repeat a non-idempotent firmware update.

zephyr/src/ota/downloader/spotflow_ota_downloader_transport_socket.c
roberthusak: Such download errors would be non-fatal, because MCUboot checks the integrity of each firmware image with its checksum (and signature). Still, it's better to detect them early on, so I added the check.

zephyr/src/ota/Kconfig
roberthusak: Good point, added `depends on !SETTINGS_NONE` as `config STREAM_FLASH_PROGRESS` [does it](https://github.com/zephyrproject-rtos/zephyr/blob/17939cff6553c5c89cbe2e3d145057068367fd39/subsys/storage/stream/Kconfig#L53).

zephyr/src/ota/core/spotflow_ota_state.c
roberthusak: AFAIK we decided not to include copyright header in each file to save space - `LICENSE.MD` in the root should be sufficient.

Therefore, I suggest that we remove this instruction from `CODING_STYLE.md` together with all copyright header remnants - `spotflow_certs.h` and some `CMakeLists.txt` files have wrong license identifiers in these headers. Let me know whether you agree with this change and I'll perform it in this PR if so.

roberthusak:
> please add the changelog and make sure all the new files conform CODING_STYLE.md mainly
> 
> * New OTA C and header files omit the required copyright header.
> * Source files generally do not include their own header first.

I added the changelog line and enforced `CODING_STYLE.md` on new and updated files. My suggestion regarding copyright headers is here: https://github.com/spotflow-io/device-sdk/pull/66#discussion_r3603046050

zephyr/include/spotflow/downloader.h
roberthusak: Good point, replaced by initialization by a registered pre-kernel callback. I also added a function for dynamic initialization.

zephyr/src/ota/downloader/spotflow_ota_downloader_transport_socket.c
```diff
@@ -0,0 +1,271 @@
...
+	if (rc < 0) {
+		LOG_ERR("Failed to connect to artifact download endpoint on port %u: %d", url->port,
+			errno);
+		zsock_close(sock);
+		return -errno;
+	}
+
+	return sock;
+}
+
```
jmasek, line 207: shouldn't we check if the state isn't paused between the callbacks?

zephyr/src/ota/downloader/spotflow_ota_downloader_transport_socket.c
```diff
@@ -0,0 +1,271 @@
...
+	if (request == NULL || request->url == NULL || request->authorization_header == NULL ||
+	    request->downloader == NULL || request->callback == NULL ||
+	    request->bytes_downloaded == NULL || request->artifact_size == NULL ||
+	    request->transient_failure == NULL) {
+		return -EINVAL;
+	}
+
+	*request->transient_failure = false;
+	*request->bytes_downloaded = 0;
+
```
jmasek, line 55: shouldn't we include also the zsock_getaddrinfo and zsock_connect into the timeout?

zephyr/src/ota/downloader/spotflow_ota_downloader.c
```diff
@@ -0,0 +1,325 @@
...
+	size_t total_bytes_downloaded = 0;
+	uint64_t artifact_size = 0;
+
+	while (true) {
+		if (downloader_is_canceled(downloader)) {
+			downloader_finish(downloader);
+			return -ECANCELED;
+		}
+
+		downloader_wait_if_paused(downloader);
```
jmasek: for transient failure, we use unbounded retries, right? is it correct?

zephyr/src/ota/core/spotflow_ota_state.c
```diff
@@ -0,0 +1,981 @@
...
+
+	if (attempt->has_attempt_error) {
+		return true;
+	}
+
+	for (size_t i = 0; i < attempt->update.artifact_count; i++) {
+		if (attempt->results[i] != SPOTFLOW_OTA_RESULT_PENDING) {
+			return true;
+		}
+	}
```
jmasek, line 1477: if main artififact is uncofirmed, then it will leave it, right? so basically you can have both cancelled status but with suceeded main firmware update, right?

zephyr/src/ota/core/spotflow_ota_state.c
jmasek: I agree, you can change the coding style in this PR and do the cleanup in seperate, it doesn't matter really.

zephyr/include/spotflow/downloader.h
```diff
@@ -0,0 +1,168 @@
...
+ * the downloader, but they are owned by the SDK and must not be modified by
+ * application code.
+ */
+struct spotflow_downloader {
+	enum spotflow_downloader_state state;
+	struct k_mutex mutex;
+	bool cancel_requested;
+	struct k_sem resume_sem;
+};
+
```
jmasek, line 42: shouldn't we document and enforce that only a single downloader is allowed. Because if there is a more downloaders, they can compete for recv_buf

zephyr/src/ota/firmware/spotflow_ota_fw_main.c
```diff
@@ -0,0 +1,802 @@
...
+	notify_main_firmware_state(&state);
+}
+
+static void notify_main_firmware_state(const struct spotflow_ota_main_firmware_state* state)
+{
+	if (state == NULL) {
+		return;
+	}
+
```
jmasek, line 492: if user/dev will use this and will call the ota api there can unbounded recursion.
spotflow_on_main_firmware_update_progressed()
  -> spotflow_get_main_firmware_update_info()
       -> spotflow_ota_init()
The second spotflow_ota_init() executes on the same thread.

There are multiple path resulting in the recursion. The option are adding another state `INITIALIZING` or do not call callback from incomplete initialization.

zephyr/src/ota/core/spotflow_ota_state.c
```diff
@@ -0,0 +1,981 @@
...
+		if (attempt_has_reportable_results(&current_attempt)) {
+			action->report_requested = true;
+		}
+		k_mutex_unlock(&state_mutex);
+		return 0;
+	}
+
+	if (attempt_has_terminal_results(&current_attempt)) {
+		start_attempt(msg, &current_attempt);
+		clear_pending_attempt();
```
jmasek: if you reboot with V1 but the check is slower than accepting the new message then you will supersede the current uncofirmed with V2, right? Then downloader starts downloading the image so corrupting the rollback image of v1, right? Effectively destroying rollback image.

zephyr/src/ota/core/spotflow_ota_worker.c
```diff
@@ -0,0 +1,355 @@
...
+	rc = spotflow_ota_state_apply_artifact_result(job->artifact_index, result, &action);
+	if (rc < 0) {
+		LOG_ERR("Failed to apply OTA result for attempt %llu artifact %zu: %d",
+			(unsigned long long)job->attempt_id, job->artifact_index, rc);
+		return rc;
+	}
+
+	spotflow_ota_state_get_snapshot(&snapshot);
+	if (snapshot.has_current_attempt && snapshot.current_attempt_id == job->attempt_id &&
+	    !snapshot.has_attempt_error) {
```
jmasek: possible race condition between Mqtt and ota worker?
spotflow_ota_state_apply_artifact_result() marks A’s artifact SUCCEEDED, clears artifact_running, and makes A terminal in RAM.
Before the worker calls spotflow_ota_state_get_snapshot(), the MQTT thread receives attempt B.
spotflow_ota_state_accept_update() sees that A is terminal and immediately calls start_attempt() for B, overwriting current_attempt.
The worker resumes and takes a snapshot, but now receives B.
The guard snapshot.current_attempt_id == job->attempt_id fails because the job belongs to A.

zephyr/src/ota/spotflow_ota.c
```diff
@@ -0,0 +1,455 @@
...
+	spotflow_ota_state_reset();
+	spotflow_ota_worker_reset();
+#if IS_ENABLED(CONFIG_SPOTFLOW_OTA_AUTO_HANDLE_MAIN_FIRMWARE)
+	spotflow_ota_fw_main_reset();
+#endif /* CONFIG_SPOTFLOW_OTA_AUTO_HANDLE_MAIN_FIRMWARE */
+}
+
+bool spotflow_is_update_canceled(void)
+{
+	if (spotflow_ota_init() < 0) {
```
jmasek, line 185: bug - checking attempt id before checking rc. Reading an uninitialized object is undefined behavior in C.

zephyr/src/ota/downloader/spotflow_ota_downloader.c
roberthusak: Yes, we do. I think it is correct, because this way the update will always eventually finish even with unstable network. If we set a bound, the user will have to manually retry all the devices that were unable to finish download within it.

If there's a request for such feature in the future, it would be easy to add. Or I can add it right away if you think it's necessary.

zephyr/docs/ota-design.md
```diff
@@ -0,0 +1,322 @@
+# Zephyr OTA updates— implementation design notes
```
jmasek: shouldn't be this file renamed? maybe just docs/ota.md? and we should reference it from the main readme

zephyr/docs/ota-design.md
```diff
@@ -0,0 +1,322 @@
...
+
+This document summarizes important design decisions on the implementation of over-the-air updates.
+
+## Module map
```
jmasek: I suggest reorganizing the document so it is more readable for users. I think you should start with exaplaining what is manifest, artifact, attempt. then put the section ## Attempt and artifact state model and then continue with less important information

README.md
```diff
@@ -157,6 +157,76 @@ end
 
 For BLE transport, the device exposes session metadata as a readable GATT characteristic and exchanges configuration data through framed TX and RX stream messages.
 
+#### OTA Updates
+
+The implementation is split into the following components:
```
jmasek, line 175: I would prefer to start with more details about implementation - states, diagrams, etc. Maybe link to the documentation - about getting started and code organization can be moved to docs folder. It is useful for user only if they want to dig into the sdk on the code level.

zephyr/samples/ota/src/main.c
```diff
@@ -0,0 +1,115 @@
...
+	case SPOTFLOW_OTA_PHASE_PENDING_REBOOT:
+		return "PENDING_REBOOT";
+	case SPOTFLOW_OTA_PHASE_UNCONFIRMED:
+		return "UNCONFIRMED";
+	default:
+		return "UNKNOWN";
+	}
+}
+
+static void confirm_unconfirmed_main_firmware(void)
```
jmasek: static definition should be after public definitions

zephyr/samples/ota/sysbuild.conf
```diff
@@ -0,0 +1 @@
+SB_CONFIG_BOOTLOADER_MCUBOOT=y
```
jmasek, line 2: can you please add explanation what this option is used?

zephyr/src/ota/downloader/spotflow_ota_downloader.c
```diff
@@ -0,0 +1,325 @@
...
+
+enum spotflow_downloader_state
+spotflow_get_downloader_state(const struct spotflow_downloader* downloader)
+{
+	if (downloader == NULL) {
+		LOG_ERR("downloader cannot be NULL");
+		return SPOTFLOW_DOWNLOADER_STATE_INACTIVE;
+	}
+
+	return downloader->state;
```
jmasek: shouldn't you hold the mutex when accessing the state? all other accesses use it.

zephyr/src/ota/downloader/spotflow_ota_downloader.c
```diff
@@ -0,0 +1,325 @@
...
+
+static void downloader_wake_waiters(struct spotflow_downloader* downloader)
+{
+	k_sem_give(&downloader->resume_sem);
+}
+
+static void downloader_drain_resume_sem(struct spotflow_downloader* downloader)
+{
+	while (k_sem_take(&downloader->resume_sem, K_NO_WAIT) == 0) {
+	}
```
jmasek: this is is a bit misleading when resume_sem is inited with capacity 1, could you add comment please?

zephyr/src/net/transport/mqtt/spotflow_mqtt.c
```diff
@@ -372,25 +438,132 @@ static void mqtt_evt_handler(struct mqtt_client* client, const struct mqtt_evt*
...
+		if (ret < 0) {
+			LOG_ERR("Failed to acknowledge MQTT PUBLISH: %d", ret);
 		}
 		break;
 	default:
 		break;
 	}
 }
 
+static int read_publish_payload(struct mqtt_client* client, uint8_t* buffer, size_t buffer_len,
```
jmasek: shouldn't be the name different - not read but something like read and discard? also I think we should log that we are discarding it.

zephyr/src/net/Kconfig.mqtt
```diff
@@ -54,4 +55,9 @@ config SPOTFLOW_INGEST_KEY
 		Ingest key is a secret key used to authenticate your device
 		It can be found in Spotflow observability platform
 
+if SPOTFLOW_OTA
+configdefault NET_SOCKETS_TLS_MAX_CONTEXTS
+	default 2
+endif # SPOTFLOW_OTA
```
jmasek, line 69: why is this needed?

zephyr/src/ota/core/spotflow_ota_log.c
```diff
@@ -0,0 +1,104 @@
...
+{
+	switch (result) {
+	case SPOTFLOW_OTA_RESULT_SUCCEEDED:
+		return "succeeded";
+	case SPOTFLOW_OTA_RESULT_FAILED:
+		return "failed";
+	case SPOTFLOW_OTA_RESULT_CANCELED:
+		return "canceled";
+	case SPOTFLOW_OTA_RESULT_PENDING:
+	default:
```
jmasek: is a good idea to use pending as default - I would prefer unknown or something that can say that something was wrong.

zephyr/src/ota/core/spotflow_ota_log.c
```diff
@@ -0,0 +1,104 @@
...
+		return "downloading";
+	case SPOTFLOW_OTA_PHASE_PENDING_UPGRADE:
+		return "pending upgrade";
+	case SPOTFLOW_OTA_PHASE_PENDING_REBOOT:
+		return "pending reboot";
+	case SPOTFLOW_OTA_PHASE_UNCONFIRMED:
+		return "unconfirmed";
+	case SPOTFLOW_OTA_PHASE_NOT_RUNNING:
+	default:
+		return "not running";
```
jmasek: is a good idea to use pending as default - I would prefer unknown or something that can say that something was wrong.

zephyr/src/ota/core/spotflow_ota_log.c
```diff
@@ -0,0 +1,104 @@
+#include "ota/core/spotflow_ota_log.h"
+
+#include <stdio.h>
+
+#include <zephyr/logging/log.h>
+
+LOG_MODULE_DECLARE(spotflow_ota, CONFIG_SPOTFLOW_MODULE_DEFAULT_LOG_LEVEL);
```
jmasek: I think that ota should have a separate ota default log level - not the spotflow one.

zephyr/src/ota/downloader/spotflow_ota_downloader.ld
```diff
@@ -0,0 +1,3 @@
+#include <zephyr/linker/iterable_sections.h>
+
+ITERABLE_SECTION_RAM(spotflow_downloader, Z_LINK_ITERABLE_SUBALIGN)
```
jmasek, line 3: why are you registering downloader this way?

zephyr/src/ota/Kconfig
```diff
@@ -0,0 +1,85 @@
+menuconfig SPOTFLOW_OTA
```
jmasek, line 1: shouldn't ota select/depends on the mcu_boot, does it make sense to have it withou mcu boot?

zephyr/include/spotflow/ota.h
```diff
@@ -0,0 +1,288 @@
...
+ *
+ * @param state Optional out parameter for the resulting snapshot.
+ *
+ * @retval 0 Update resumed.
+ * @retval -EINVAL No active main-firmware update, or the update is not paused.
+ * @retval -ENOTSUP Automatic main-firmware handling is disabled.
+ * @retval <0 Negative errno when OTA initialization or the underlying download
+ *         resume fails.
+ */
+int spotflow_resume_main_firmware_update(struct spotflow_ota_main_firmware_state* state);
```
jmasek, line 225: is this used anywhere else than in the test, is it intentional?

zephyr/src/ota/spotflow_ota.c
```diff
@@ -0,0 +1,455 @@
...
+
+int spotflow_resume_main_firmware_update(struct spotflow_ota_main_firmware_state* state)
+{
+#if !IS_ENABLED(CONFIG_SPOTFLOW_OTA_AUTO_HANDLE_MAIN_FIRMWARE)
+	if (state != NULL) {
+		(void)spotflow_get_main_firmware_update_state(state);
+	}
+
+	LOG_ERR("Main firmware auto-handling is not enabled");
+	return -ENOTSUP;
```
jmasek: why is this called, when -ENOTSUP is called anyway?

zephyr/src/ota/spotflow_ota.c
```diff
@@ -0,0 +1,455 @@
...
+
+int spotflow_abort_main_firmware_update(struct spotflow_ota_main_firmware_state* state)
+{
+#if !IS_ENABLED(CONFIG_SPOTFLOW_OTA_AUTO_HANDLE_MAIN_FIRMWARE)
+	if (state != NULL) {
+		(void)spotflow_get_main_firmware_update_state(state);
+	}
+
+	LOG_ERR("Main firmware auto-handling is not enabled");
+	return -ENOTSUP;
```
jmasek: same as above - why is this called, when -ENOTSUP is called anyway? Btw. this is general ota file or auto handle main firmware :D because I am a bit confused because this seems to be for the auto handle main firmware

zephyr/src/ota/core/spotflow_ota_state.c
roberthusak: `zephyr/docs/ota.md` should cover this now.

zephyr/src/ota/core/spotflow_ota_state.c
roberthusak: Good point. Reworked the state-handling logic and explicitly implemented this case as "rehydration".

zephyr/src/ota/core/spotflow_ota_state.c
roberthusak: The worker is redone to handle all these cases, especially retry transient persistence errors.

zephyr/docs/ota-design.md
roberthusak: I agree, renamed. It's already referenced from the bottom of the OTA section in the main README.

zephyr/docs/ota-design.md
roberthusak: Reorganized. It starts with an overview (containing a glossary), continues with top-level design (such as those state machines), and ends with implementation info (threads, synchronization between worker and state, etc.).

README.md
roberthusak: I updated this section to be standalone but short for quick overview. The reader can still read `ota.md` for more design and implementation details.

I kept the module diagram because it explains parts of the system used in the sequence diagram. However, I'd rather not duplicate state diagrams here. If you think the current description is insufficient, I suggest we remove both diagrams (including the component list) and just keep the first two paragraphs and the links.

zephyr/src/ota/core/spotflow_ota_state.c
roberthusak: OK, removed.

zephyr/src/ota/downloader/spotflow_ota_downloader_transport_socket.c
roberthusak: We should, fixed.

zephyr/src/ota/downloader/spotflow_ota_downloader_transport_socket.c
roberthusak: We should, fixed.

zephyr/src/ota/core/spotflow_ota_state.c
roberthusak: I'm not sure how you mean this question in the context of this function (leave what?). However, in general, it's possible that one artifact update (such as the main firmware) results in `SUCCEEDED` while another one in `CANCELED`. AFAIK the portal then classifies the whole device update attempt as `CANCELED`. However, the device doesn't send anything like a cancellation confirmation to the cloud, just the results of the individual artifacts (OR an error of the whole attempt in case it's malformed etc., but that's evaluated before processing the artifacts).

Regarding main firmware update, once the image is upgraded, receiving a cancellation request from the cloud doesn't cancel it anymore. In that case, the SDK waits for the main firmware update to finish. And if it finishes successfully and there are more artifacts after it, they won't be canceled either to prevent half-finished updates.

zephyr/src/ota/core/spotflow_ota_worker.c
roberthusak: The reworked mechanism (mutations, staging, etc.) should now guard against cases like these.

zephyr/src/ota/spotflow_ota.c
roberthusak: This function (`spotflow_is_update_canceled`) doesn't check the attempt ID. Did you have another one in mind?

zephyr/src/ota/core/spotflow_ota_log.c
roberthusak: Good point, fixed and added a warning when that happens.

zephyr/src/ota/core/spotflow_ota_log.c
roberthusak: Good point, fixed and added a warning when that happens.

zephyr/src/net/Kconfig.mqtt
roberthusak: Added a comment: One socket is needed for the MQTT connection and the other one for downloading artifacts.

zephyr/samples/ota/sysbuild.conf
roberthusak: Added.

zephyr/samples/ota/src/main.c
roberthusak: Fixed coding style of the file.

zephyr/src/ota/downloader/spotflow_ota_downloader.ld
roberthusak: The structure `spotflow_downloader` contains a mutex and a semaphore that must be initialized with `k_mutex_init` and `k_sem_init` before usage. While the user code can call `spotflow_init_downloader` to perform the initialization dynamically, we also provide `SPOTFLOW_DEFINE_DOWNLOADER` for the global code - in the same way as Zephyr mutexes have both `k_mutex_init` and `K_MUTEX_DEFINE`.

`SPOTFLOW_DEFINE_DOWNLOADER` uses `ITERABLE_SECTION_RAM` and `STRUCT_SECTION_ITERABLE` so that it can initialize all globally defined downloaders with `init_static_downloaders`.

zephyr/src/ota/core/spotflow_ota_log.c
roberthusak: OK, added.

zephyr/src/ota/Kconfig
roberthusak: Only automatic handling of main firmware updates (which is enabled by default though) depends on MCUboot.

zephyr/include/spotflow/ota.h
roberthusak: I didn't want to make the sample complicated by including it. It's a part of the public API and should work as expected though. Its implementation `spotflow_ota_fw_main_resume_update` is covered by tests.

zephyr/src/ota/spotflow_ota.c
roberthusak: I think the original idea was that `state` (when non-null) should receive a valid value even when the return code is <0, but I don't like it either 😅 Do you think it's sufficient to print the error and return `-ENOTSUP`? Or would it be better not to compile this function at all when `SPOTFLOW_OTA_AUTO_HANDLE_MAIN_FIRMWARE` isn't enabled?

zephyr/src/ota/spotflow_ota.c
roberthusak: dtto (there are multiple functions like these)

zephyr/include/spotflow/downloader.h
roberthusak: Moved the receive buffer to the downloader instance so that this isn't necessary.

zephyr/src/ota/downloader/spotflow_ota_downloader.c
roberthusak: Good point, updated.

zephyr/src/ota/downloader/spotflow_ota_downloader.c
roberthusak: You're right, a loop isn't necessary. Simplified.

zephyr/src/net/transport/mqtt/spotflow_mqtt.c
roberthusak: Makes sense - renamed and added logging to the remaining call site. Also made the read blocking to ensure the whole payload is drained before another message is processed.

zephyr/src/ota/core/spotflow_ota_state.c
roberthusak: This shouldn't happen now (at least after the latest updates to state handling) - even supersession must wait before the update of the current artifact finishes, including the update of the main firmware.

Still, I rather added an explicit check that the current image is confirmed into `spotflow_ota_platform_begin_image_write`.

zephyr/src/ota/firmware/spotflow_ota_fw_main.c
roberthusak: Good point, fixed - callbacks are now deferred after the initialization.

zephyr/samples/ota/prj.conf
roberthusak: Eventually fixed in this PR.

roberthusak:
> do you consider adding automatic pipeline build of ota sample for some boards?

I'll do that in a separate PR if that's OK. I think we should do it for all the samples, now we build only `logs` in CI.

jmasek:
> > do you consider adding automatic pipeline build of ota sample for some boards?
> 
> I'll do that in a separate PR if that's OK. I think we should do it for all the samples, now we build only `logs` in CI.

sure

zephyr/src/ota/spotflow_ota.c
```diff
@@ -0,0 +1,449 @@
...
+
+static void handle_ota_c2d_msg(uint8_t* payload, size_t len)
+{
+	static struct spotflow_ota_cbor_c2d_msg msg;
+	struct spotflow_ota_cbor_decode_status status;
+	int rc = spotflow_ota_cbor_decode_c2d(payload, len, &msg, &status);
+
+	if (status.has_trustworthy_attempt_id) {
+		update_last_received_attempt_id(status.attempt_id);
+	}
```
jmasek, line 293: wrong place in previous comment - bug - checking attempt id before checking rc. Reading an uninitialized object is undefined behavior in C.

zephyr/src/ota/spotflow_ota.c
jmasek: the original is in spotflow_ota.c:326-330, added comment there

jmasek (changes requested):

zephyr/src/ota/spotflow_ota.c
roberthusak: Fixed by zeroing `status` before passing it to `spotflow_ota_cbor_decode_c2d`.

In practice, `spotflow_ota_cbor_decode_c2d` would zero `status` every time, but it's indeed better not to have this hidden contract.

zephyr/docs/ota.md
```diff
@@ -0,0 +1,781 @@
...
+```mermaid
+flowchart TD
+    cloud["Cloud sends UPDATE_ARTIFACTS<br/>with attempt ID and ordered manifest"]
+    accept["Validate and accept attempt"]
+    persistAttempt["Persist accepted attempt"]
+    next["Select next pending artifact"]
+    installed{"Requested version<br/>already installed?"}
+    kind{"Automatic<br/>main firmware?"}
+    skip["Record SUCCEEDED"]
+    delegated["Invoke delegated<br/>application handler"]
```
jmasek, line 80: I do not understand when succeeded it ends up in the Record terminal result and not Succeeded.

roberthusak: "Record SUCCEEDED" is just a special case of "Record terminal result". The terminal result can be either SUCCEEDED, FAILED, or CANCELED, depending on the value returned from the handler.

The transition from "Invoke delegated application handler" goes just to "Record terminal result" so that we don't need to put another decision node (which would inspect the value returned from the handler) to the flowchart.

I don't know how to modify the flowchart to make this thing more readable. Should I add the second line "(SUCCEEDED, FAILED, CANCELED)" to "Record terminal result"?

jmasek (changes requested):

zephyr/docs/ota.md
jmasek: yes, I think it is a good idea - because succeed is terminal result right. so I would put suceeded, failde, canceled and omit terminal result completely

zephyr/src/ota/spotflow_ota.c
jmasek: I do not know - it is just misleading - you call update state but no matter what you call not supported error, it is quite strong side effect. If it is not supported, I would consider preventing user from calling it.

roberthusak: Yeah, I agree. Both the declarations and definitions are now guarded by `CONFIG_SPOTFLOW_OTA_AUTO_HANDLE_MAIN_FIRMWARE` so that the code using it doesn't even compile when the option isn't enabled.

zephyr/docs/ota.md
roberthusak: OK, done. I still kept "Record SUCCEEDED" separate as well so that it's apparent that it's reported when the firmware version is already installed and after the image is confirmed.

jmasek: now you have succeeded twice there, it is not correct, right? result["Record SUCCEEDED, FAILED, or CANCELED"]

roberthusak: Each result is a separate node now.

jmasek (approved):

PR #67, opened 2026-02-26
Added coredump checksum check.

noorhaq:
Added the coredump checksum check to verify coredumps before uplaods.


esp_idf/spotflow/device_sdk/src/coredump/spotflow_coredump.c
```diff
@@ -30,6 +30,14 @@ typedef struct {
 
 static coredump_info_t coredump_info = { 0 };
 
+/**
+ * @brief Check if coredump is valid by checking it's checksum
+ *
+ * @return true it is valud
+ * @return false it is not valid sending error message to cloud
+ */
+esp_err_t spotflow_is_coredump_valid(void);
```
roberthusak: Shouldn't we make it static? Also, should the comment be present both on the declaration and implementation?

esp_idf/spotflow/device_sdk/src/coredump/spotflow_coredump.c
```diff
@@ -97,6 +105,11 @@ esp_err_t spotflow_coredump_backend(void)
 		return ESP_ERR_NOT_FOUND;
 	}
 
+	err = spotflow_is_coredump_valid();
+	if (err != ESP_OK) {
+		SPOTFLOW_LOG("Coredump is invalid, will attempt to send error report to cloud.");
```
roberthusak: The comment seems misleading - the report can be already sent to the cloud by now, right?

esp_idf/spotflow/device_sdk/src/coredump/spotflow_coredump.c
```diff
@@ -211,4 +224,61 @@ esp_err_t spotflow_coredump_backend(void)
...
\ No newline at end of file
+}
+
+/**
+ * @brief Check if coredump is valid by checking it's checksum
+ *
+ * @return true it is valud
+ * @return false it is not valid sending error message to cloud
+ */
+esp_err_t spotflow_is_coredump_valid(void)
```
roberthusak: The name of the function seems misleading - it only checks the image validity on the start, then it consists mostly of the same sending logic as `spotflow_coredump_backend`. Would it make sense to move `esp_core_dump_image_check()` to `spotflow_coredump_backend` and let this function (after renaming) just send the empty chunk?

Also, I'd consider if there are any parts of this function that can be shared with the sending logic in `spotflow_coredump_backend`. Wouldn't be even possible to eliminate an additional function altogether by just setting `coredump_info.size = 0` in `spotflow_coredump_backend` if the checksum is invalid?

roberthusak (commented):
The functionality seems good 👍 Just added some comments regarding code style.

esp_idf/spotflow/device_sdk/src/coredump/spotflow_coredump.c
noorhaq: Yep tried and build it like this.

esp_idf/spotflow/device_sdk/src/coredump/spotflow_coredump.c
```diff
@@ -97,6 +97,13 @@ esp_err_t spotflow_coredump_backend(void)
 		return ESP_ERR_NOT_FOUND;
 	}
 
+	err = esp_core_dump_image_check();
+	if (err != ESP_OK) {
+		SPOTFLOW_LOG("Coredump is invalid, will attempt to send error report to cloud.");
+		ESP_LOGE(TAG, "Coredump is invalid: checksum mismatch.");
+		coredump_size = 1; // Set size to 1 to indicate invalid coredump, actual content will be empty
```
roberthusak: Will the content really be empty? I'd expect that a coredump with just one byte would be send (which would be better to avoid).

esp_idf/spotflow/device_sdk/src/coredump/spotflow_coredump.c
```diff
@@ -188,11 +198,12 @@ esp_err_t spotflow_coredump_backend(void)
 		// Update progress
 		coredump_info.chunk_ordinal++;
 		coredump_info.offset += current_chunk_size;
-
+		if (coredump_info.size > 0) {
```
roberthusak, line 201: I think we should log even when an empty chunk (i.e., due to core dump corruption) was sent, so there should probably be an `else` branch, too.

Nitpick: Please indent the command nested in the block.

noorhaq: Done.

roberthusak (approved):

PR #72, opened 2026-04-10
Add support for Zephyr 4.4.0

roberthusak:


zephyr/ci/boards.yml
```diff
@@ -243,11 +248,6 @@ vendors:
       - id: rpi_pico_w
         name: Raspberry Pi Pico W
         board: rpi_pico/rp2040/w
-        manifest: west-zephyr-raspberrypi-pico-w.yml
```
jmasek, line 246: so this was fixed in 4.4?

zephyr/samples/coredumps/boards/esp32c6_devkitc_esp32c6_hpcore.overlay
jmasek, line 1: does it mean that we should build with different target or can we keep devkitc target (-b parameter for west build)

zephyr/Kconfig
```diff
@@ -24,6 +23,22 @@ menuconfig SPOTFLOW
...
+	select PSA_WANT_ALG_GCM if !NRF_SECURITY
+	select PSA_WANT_ALG_HMAC if !NRF_SECURITY
+	select PSA_WANT_ALG_RSA_PKCS1V15_SIGN if !NRF_SECURITY
+	select PSA_WANT_ALG_SHA_384 if !NRF_SECURITY
+	select PSA_WANT_ALG_TLS12_PRF if !NRF_SECURITY
+	select PSA_WANT_ECC_SECP_R1_256 if !NRF_SECURITY
+	select PSA_WANT_KEY_TYPE_AES if !NRF_SECURITY
+	select PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_EXPORT if !NRF_SECURITY
+	select PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_GENERATE if !NRF_SECURITY
+	select PSA_WANT_KEY_TYPE_RSA_KEY_PAIR_IMPORT if !NRF_SECURITY
```
jmasek, line 41: are these options backwards compatible? Will it work with Zephyr 4.3, 4.2, etc?

zephyr/ci/boards.yml
roberthusak: Yes, I tried it and it worked 👍

zephyr/Kconfig
roberthusak: Yes - it compiles for Zephyr 4.3 and 3.7. And it's tested to work on 4.3.

zephyr/samples/coredumps/boards/esp32c6_devkitc_esp32c6_hpcore.overlay
roberthusak: The problem is that `esp32c6_devkitc` is not a valid target since Zephyr 4.2 because we need to choose which core (`hpcore` or `lpcore`) to build for. Therefore, `esp32c6_devkitc.conf` and `esp32c6_devkitc.overlay` are wrong file names because they don't match the full target name (hence they won't be applied).

We've already fixed this issue in other samples, but the `coredumps` sample went through the cracks until now.

jmasek (approved):

PR #73, opened 2026-04-12
Feature/esp idf metrics

noorhaq:
It is working but due to some reason following 2 things are reported as dropped logs too many.
And I think the dashboard is updating very late. Like for my device qemu. It is updating every few seconds. But dashboard updates after like 10-20 mins with old values. 
Not sure on these because as per the esp idf logs the data is being sent to cloud. Maybe malformed or something if we can confirm with backend.


noorhaq:
Logs example is fixed in the coredump hash PR.

esp_idf/spotflow/device_sdk/examples/metrics/main/app_main.c
```diff
@@ -0,0 +1,218 @@
...
+		// 	ESP_LOGD(TAG, "Debug log message works");
+		// 	ESP_LOGE(TAG, "Error log message works");
+		// 	ESP_LOGW(TAG, "Warning log message works");
+		// 	ESP_LOGV(TAG, "Verbose log message works");
+
+		// 	//Testing long log messages. With differenet types
+		// 	ESP_LOGI(TAG,
+		// 		 "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789abc"
+		// 		 "defghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789abcdef"
+		// 		 "ghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789abcdefghi"
```
jmasek: Please remove commented-out log messages.

esp_idf/spotflow/device_sdk/examples/metrics/main/app_main.c
```diff
@@ -0,0 +1,218 @@
...
+static void report_counter_metric(void)
+{
+    static int counter = 0;
+    counter += 10;
+
+    int rc = spotflow_report_metric_int(g_app_counter_metric, counter);
+    if (rc < 0) {
+        // ESP_LOGI(TAG,"Failed to report counter metric: %d", rc);
+    } else {
+        // ESP_LOGD(TAG,"Reported counter: %d", counter);
```
jmasek: why commented out?

esp_idf/spotflow/device_sdk/examples/metrics/main/app_main.c
```diff
@@ -0,0 +1,218 @@
...
+		// 		 "jklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789abcdefghijkl"
+		// 		 "mnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789abcdefghijklmno"
+		// 		 "pqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789abcdefghijklmnopqr"
+		// 		 "stuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789");
+
+		// 	ESP_LOGW(TAG, "Test Number. %d", test_num++);
+		// 	ESP_LOGI(TAG, "[APP] Free memory: %" PRIu32 " bytes",
+		// 		 esp_get_free_heap_size());
+		// }
+		vTaskDelay(50000/portTICK_PERIOD_MS); // Delay for 50s
```
jmasek: So if the reporting thread fails, the main will continue to operate. Normally, is it a standard/good idea?

esp_idf/spotflow/device_sdk/examples/metrics/partitions.csv
```diff
@@ -0,0 +1,6 @@
+# Name,   Type, SubType, Offset,  Size
+# Note: if you have increased the bootloader size, make sure to update the offsets to avoid overlap
+nvs,      data, nvs,     0x9000,  0x6000
+phy_init, data, phy,     0xf000,  0x1000
+factory,  app,  factory, 0x10000, 0x150000
+coredump, data, coredump,,        256K
```
jmasek: why are you changing partitions? Does this example also include the coredump example?

esp_idf/spotflow/device_sdk/examples/metrics/main/app_main.c
noorhaq: i was just testing at the time like should we keep the normal logs example kind of with metrics or not?
i have un commented.

esp_idf/spotflow/device_sdk/examples/metrics/main/app_main.c
noorhaq: If the reporting thread fails the main will continue as reporting is like a low priority standard like we don't want the device to stop working or lagging if the reporting is covering much memory. So even if the reporting thread fails the main and other threads shouldn't be effected.

esp_idf/spotflow/device_sdk/examples/metrics/partitions.csv
noorhaq: I have kept it here, so coredump can be enabled as well and could do some regression testing i.e. metrics. logs and coredump all 3 can work simultaneously. Like I can remove it if required? because for coredump we don't actually need anything just need to enable like a config.

jmasek: honestly, I would prefer the same hierarchy as zephyr - logs - only logs, metrics - logs+metrics, coredumps - complete example - coredumps+logs+metrics.

esp_idf/spotflow/device_sdk/src/metrics/spotflow_metrics_net.c
```diff
@@ -0,0 +1,98 @@
...
+	if (pub_rc < 0) {
+		SPOTFLOW_LOG("Failed to publish metric: %d", pub_rc);
+		return pub_rc; /* Message stays in queue. Retry later */
+	}
+
+	/* Only remove after successful publish */
+	xQueueReceive(g_spotflow_metrics_msgq, &msg, 0);
+
+	free(msg->payload);
+	free(msg);
```
jmasek, line 66: you are sending only a single metric message per SPOTFLOW_MQTT_NOTIFY_METRICS wakeup, then always clears the event bit. Event-group bits are not counted, so if a collection cycle enqueues several metric payloads at once, all but the first can sit in the queue until some later notification happens.

esp_idf/spotflow/device_sdk/src/metrics/spotflow_metrics_aggregator.c
```diff
@@ -0,0 +1,347 @@
...
+
+	int rc = spotflow_metrics_cbor_encode_no_aggregation(
+	    metric, labels, label_count, value_int, value_float, esp_timer_get_time() / 1000ULL,
+	    seq_num, &cbor_data, &cbor_len);
+	if (rc < 0)
+		return rc;
+
+	rc = spotflow_metrics_enqueue(cbor_data, cbor_len);
+	if (rc < 0)
+		free(cbor_data);
```
jmasek: but when enqueue is successful, who frees the cbor_data? because spotflow_metrics_enqueue is not freeing it.

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system_network.c
```diff
@@ -0,0 +1,178 @@
...
+ */
+void spotflow_metrics_system_network_collect(void)
+{
+	if (!g_network_tx_metric || !g_network_rx_metric) {
+		SPOTFLOW_LOG("Network metrics not registered");
+		return;
+	}
+
+	// Iterate over all interfaces
+	for (int i = 0; i < esp_netif_get_nr_of_ifs(); i++) {
```
jmasek: is there any check to prevent i >= CONFIG_SPOTFLOW_METRICS_SYSTEM_NETWORK_MAX_INTERFACES? so we won't be accessing g_hooks outside the array?

esp_idf/spotflow/device_sdk/spotflow.c
```diff
@@ -17,8 +17,30 @@
...
 #include "configs/spotflow_config.h"
 
+#ifdef CONFIG_SPOTFLOW_METRICS
+#include "metrics/spotflow_metrics_registry.h"
+#include "metrics/spotflow_metrics_net.h"
+#endif
+
+#ifdef CONFIG_SPOTFLOW_METRICS_HEARTBEAT
+#include "metrics/spotflow_metrics_heartbeat.h"
+#include "metrics/system/spotflow_metrics_system.h"
```
jmasek, line 29: why system metric is gated by heartbeat?

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system.c
```diff
@@ -0,0 +1,164 @@
...
+static void collection_timer_handler(void* arg);
+
+int spotflow_metrics_system_init(void)
+{
+	if (atomic_load(&g_system_metrics_init_state) == 2) {
+		return 0;
+	}
+
+	int expected = 0;
+	if (!atomic_compare_exchange_strong(&g_system_metrics_init_state, &expected, 1)) {
```
jmasek: if any subsystem init fails, g_system_metrics_init_state stays at 1. Every later caller then spins forever waiting for state 2, so one transient init failure can permanently deadlock retries. Is it correct?

esp_idf/spotflow/device_sdk/CMakeLists.txt
```diff
@@ -1,46 +1,129 @@
...
+# ---------------------------------------------------------------------------
+if(CONFIG_SPOTFLOW_METRICS)
+    file(GLOB METRICS_SRCS CONFIGURE_DEPENDS "src/metrics/*.c")
+    file(GLOB METRICS_SYSTEM_SRCS CONFIGURE_DEPENDS "src/metrics/system/*.c")
+
+    # -- Prune disabled system metric sources --
+    if(NOT CONFIG_SPOTFLOW_METRICS_SYSTEM_RESET_CAUSE)
+        list(REMOVE_ITEM METRICS_SYSTEM_SRCS
+            "${CMAKE_CURRENT_SOURCE_DIR}/src/metrics/system/spotflow_reset_helper.c"
+            "${CMAKE_CURRENT_SOURCE_DIR}/src/metrics/system/spotflow_metrics_system.c"
```
jmasek: is this intended? you are removing system metrics when reset cause is not used?

esp_idf/spotflow/device_sdk/include/metrics/system/spotflow_metrics_system.h
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * @brief Report MQTT connection state change
+ *
+ * Called by MQTT layer when connection state changes.
+ * Immediately reports connection state as event metric.
+ *
+ * @param connected True if connected, false if disconnected
+ */
+void spotflow_metrics_system_report_connection_state(bool connected);
```
jmasek, line 56: this is not used, right?

esp_idf/spotflow/device_sdk/src/metrics/spotflow_metrics_heartbeat.c
```diff
@@ -0,0 +1,138 @@
...
+	if (!g_heartbeat_mutex) {
+		SPOTFLOW_LOG("Failed to create heartbeat mutex");
+		return;
+	}
+
+	const esp_timer_create_args_t timer_args = { .callback = &heartbeat_timer_callback,
+						     .arg = NULL,
+						     .name = "heartbeat_timer" };
+
+	ESP_ERROR_CHECK(esp_timer_create(&timer_args, &g_heartbeat_timer));
```
jmasek, line 77: couldn't you send the first heartbeat right after setup?

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system_cpu.c
```diff
@@ -0,0 +1,86 @@
...
+ * @brief Get CPU utilization
+ *
+ * @return float
+ */
+static float get_cpu_utilization()
+{
+	uint32_t idle0_start = get_idle_runtime();
+	uint32_t total_start = esp_timer_get_time(); // microseconds
+
+	vTaskDelay(pdMS_TO_TICKS(1000)); // sample window = 1 sec
```
jmasek, line 34: is this a good idea - it seems that this will run from an esp_timer callback so it can negativelly affect other calls running on timer callback.

esp_idf/spotflow/device_sdk/Kconfig.metrics_systems
```diff
@@ -0,0 +1,109 @@
+menuconfig SPOTFLOW_METRICS_SYSTEM
+	bool "Enable Spotflow system metrics auto-collection"
```
jmasek, line 2: should depend on metrics

esp_idf/spotflow/device_sdk/src/metrics/spotflow_metrics_net.c
```diff
@@ -0,0 +1,98 @@
...
+		return -ENOMEM;
+	}
+
+	memcpy(msg->payload, payload, len);
+	msg->len = len;
+
+	if (xQueueSend(g_spotflow_metrics_msgq, &msg, 0) != pdTRUE) {
+		free(msg->payload);
+		free(msg);
+		return -EAGAIN;
```
jmasek: shouldn't  be the ENOBUFS be used?

jmasek (changes requested):

esp_idf/spotflow/device_sdk/src/metrics/spotflow_metrics_net.c
noorhaq: Nope it remains their until the metric queue is empty. Like how it is working is.
Every time the spotflow metrics enqueue is called even on the first one it adds the msg to the queue and event bit is set to 1. If the queue is filled again the event bit remains one. It only goes to 0 or unset by the spotflow_net.c when the event queue is empty.

esp_idf/spotflow/device_sdk/src/metrics/spotflow_metrics_heartbeat.c
noorhaq: Timer callback already calls it for the first time. Like it was sending duplicates. If we send again during setup.

esp_idf/spotflow/device_sdk/src/metrics/spotflow_metrics_net.c
noorhaq: That could be confusing because we will be trying again after the queue is freed. So, if we use ENOBUF it will mean we have run out of buffer space and cause confusion.

esp_idf/spotflow/device_sdk/CMakeLists.txt
noorhaq: Yep this was un intentional. Like it was supposed to be in the else block. I will update it.

esp_idf/spotflow/device_sdk/include/metrics/system/spotflow_metrics_system.h
noorhaq: Yes because I wasn't sure how will it report the connection state if it is not connected. Like if connected it will always be true. When it is false we won't be able to send.

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system.c
noorhaq: Yes but it will only be the metrics tasks. Like if the metrics system is not able to respond. Like it kinds of wait for it to be active.

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system_network.c
noorhaq: I will add this. Forgot to add this check.

esp_idf/spotflow/device_sdk/src/metrics/spotflow_metrics_aggregator.c
noorhaq: I will check because I think I freed it somewhere.

esp_idf/spotflow/device_sdk/Kconfig.metrics_systems
noorhaq: This option won't be displayed until the metrics are enabled like kind of nested approach.

esp_idf/spotflow/device_sdk/spotflow.c
noorhaq: I had added a separate check on this I think it got reverted in some commits. And I forgot to add the change. Like it was in the metrics check. And heartbeat was nested inside that change.

esp_idf/spotflow/device_sdk/CMakeLists.txt
```diff
@@ -1,46 +1,129 @@
...
+    "src/configs/*.c"
+    "spotflow.c"
+)
+
+set(INCLUDE_DIRS
+    "include/net"
+    "include/logging"
+    "include/configs"
+    "include"
 )
```
roberthusak, line 15: `"include/metrics"` is missing. However, in general, isn't just `"include"` sufficient? It seems to me that all headers are included with their full relative paths anyway (e.g., `#include "logging/spotflow_log_backend.h"`).

If we decide to use just `"include"`, also all the following `list(APPEND INCLUDE_DIRS "include/...")` should be removed.

esp_idf/spotflow/device_sdk/examples/metrics/main/app_main.c
jmasek: I meant for example purposes. However, most likely users will use it same way, so yes, you are right

esp_idf/spotflow/device_sdk/Kconfig.metrics_systems
jmasek: why it won't be displayed? import from KConfig.metrics is not inside the if metrics and dependency on metrics is not stated anywhere. I also checked the menuconfig and it is indeed displayed.

esp_idf/spotflow/device_sdk/CMakeLists.txt
```diff
@@ -1,46 +1,129 @@
...
+        )
+    endif()
+
+    if(NOT CONFIG_SPOTFLOW_METRICS_SYSTEM_NETWORK)
+        list(REMOVE_ITEM METRICS_SYSTEM_SRCS
+            "${CMAKE_CURRENT_SOURCE_DIR}/src/metrics/system/spotflow_metrics_system_network.c"
+        )
+    else()
+        idf_build_set_property(COMPILE_DEFINITIONS
+            "LWIP_SNMP=1;MIB2_STATS=1;LWIP_NETIF_EXT_STATUS_CALLBACK=1"
```
jmasek: can you please add the explanation why

esp_idf/spotflow/device_sdk/src/net/spotflow_mqtt.c
```diff
@@ -13,7 +13,9 @@
 #ifdef CONFIG_ESP_COREDUMP_ENABLE
 #include "coredump/spotflow_coredump_net.h"
 #endif
```
jmasek, line 15: is it possible to use spotflow metrics without logs?

esp_idf/spotflow/device_sdk/src/metrics/spotflow_metrics_cbor.c
```diff
@@ -0,0 +1,339 @@
...
+                                     struct metric_timeseries_state* ts);
+
+#define CBOR_CHECK(expr)        \
+    do {                        \
+        if ((err = (expr)) != CborNoError) \
+            goto fail;          \
+    } while (0)
+
+/**
+ * @brief Debugging Function not to be used in Production
```
roberthusak: But now it's still called even in production, right? I think it should be either compiled conditionally, or removed completely. I'd prefer the latter, we can then have unit tests for checking the resulting CBOR format and contents. And an attached debugger displaying the buffer contents can still be used when debugging a particular issue.

esp_idf/spotflow/device_sdk/src/net/spotflow_mqtt.c
```diff
@@ -180,6 +182,12 @@ void spotflow_mqtt_publish(void* pvParameters)
 				clear_mask |= SPOTFLOW_MQTT_NOTIFY_LOGS;
 				}
```
jmasek, line 202: Metrics should have higher priority than logs

esp_idf/spotflow/device_sdk/src/metrics/spotflow_metrics_cbor.c
```diff
@@ -0,0 +1,339 @@
...
+	if (cbor_encoder_close_container(&encoder, &map) != CborNoError) {
+        free(buffer);
+        return -EINVAL;
+    }
+
+	size_t len = cbor_encoder_get_buffer_size(&encoder, buffer);
+    *cbor_data = malloc(len);
+    if (!*cbor_data) {
+        free(buffer);
+        return -12;
```
roberthusak:
```suggestion
        return -ENOMEM;
```

esp_idf/spotflow/device_sdk/include/metrics/spotflow_metrics_registry.h
```diff
@@ -0,0 +1,109 @@
+#ifndef SPOTFLOW_METRICS_REGISTRY_H_
+#define SPOTFLOW_METRICS_REGISTRY_H_
+
+#include "spotflow_metrics_types.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+void spotflow_metrics_init(void);
```
roberthusak, line 14: This function should be documented (in consistency with the other ones in the file). At least, the user should be dicouraged from calling it manually, as it's already done by Spotflow.

esp_idf/spotflow/device_sdk/src/metrics/spotflow_metrics_aggregator.c
```diff
@@ -0,0 +1,347 @@
...
+			int rc = flush_timeseries(metric, ts, timestamp_ms);
+			if (rc < 0)
+				SPOTFLOW_LOG("Failed to flush timeseries for metric '%s': %d",
+					     metric->name, rc);
+		}
+	}
+
+	uint32_t interval_ms = get_interval_ms(metric->agg_interval);
+	if (interval_ms > 0)
+		esp_timer_start_once(ctx->aggregation_timer, interval_ms * 1000ULL);
```
jmasek: shouldn't we check return code?

esp_idf/spotflow/device_sdk/src/metrics/spotflow_metrics_heartbeat.c
```diff
@@ -0,0 +1,138 @@
...
+		if (retry < 3) {
+			SPOTFLOW_DEBUG("MQTT busy, retrying heartbeat in %d ms...",
+				       retry_delays_ms[retry]);
+			vTaskDelay(pdMS_TO_TICKS(retry_delays_ms[retry]));
+		}
+	}
+
+	/* Always free message memory */
+	free(msg->payload);
+	free(msg);
```
jmasek: shouldn't you set g_pending_heartbeat to null when you freed it?

esp_idf/spotflow/device_sdk/src/metrics/spotflow_metrics_registry.c
roberthusak, line 1: This file uses spaces instead of tabs unlike the rest of the code.

esp_idf/spotflow/device_sdk/src/metrics/spotflow_metrics_net.c
jmasek: wtf? spotflow_poll_and_process_enqueued_metrics processes one metrics message, right?
and the caller code looks like this - it means you send one message and clear bit!

#ifdef CONFIG_SPOTFLOW_METRICS
			if (notify_value & SPOTFLOW_MQTT_NOTIFY_METRICS) {
				spotflow_poll_and_process_enqueued_metrics();
				clear_mask |= SPOTFLOW_MQTT_NOTIFY_METRICS;
			}
#endif
			xEventGroupClearBits(spotflow_mqtt_event_group, clear_mask);

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system_stack.c
```diff
@@ -0,0 +1,298 @@
...
+ */
+static void report_thread_stack(TaskHandle_t thread, void* user_data)
+{
+	(void)user_data;
+
+	if (thread == NULL) {
+		return;
+	}
+
+	UBaseType_t unused_bytes = uxTaskGetStackHighWaterMark(thread);
```
jmasek: function returns words not bytes

esp_idf/spotflow/device_sdk/include/metrics/spotflow_metrics_net.h
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * Dequeues one message from the metrics queue and publishes via MQTT.
+ * Called repeatedly by processor thread to drain the queue.
+ *
+ * @return 1 if message processed, 0 if queue empty, negative errno on failure
+ */
+int spotflow_poll_and_process_enqueued_metrics(void);
+
+/**
+ * @brief Create the enqueu metric messages
```
roberthusak:
```suggestion
 * @brief Enqueue a metric message for sending.
```

esp_idf/spotflow/device_sdk/src/metrics/spotflow_metrics_net.c
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	msg->payload = malloc(len);
+	if (!msg->payload) {
+		free(msg);
+		return -ENOMEM;
+	}
+
+	memcpy(msg->payload, payload, len);
+	msg->len = len;
```
roberthusak: Is the alocation and copying necessary? Can't we just pass `payload` directly to `msg->payload`? Would solve the comment https://github.com/spotflow-io/device-sdk/pull/73#discussion_r3130723775 as well (it will be deallocated when the message is processed). 

If you remove the extra allocation, the information about passing the ownership of the pointer `payload` should be added to the comment of this function in the header.

esp_idf/spotflow/device_sdk/src/metrics/spotflow_metrics_backend.c
```diff
@@ -0,0 +1,212 @@
...
+static int validate_labels(const struct spotflow_metric_base* base,
+			   const struct spotflow_label* labels, uint8_t label_count);
+
+/**
+ * @brief Report an integer metric value
+ *
+ * @param metric
+ * @param value
+ * @return int
+ */
```
roberthusak: Remove these comments from the source file, they are already present in the header.

esp_idf/spotflow/device_sdk/include/metrics/spotflow_metrics_aggregator.h
```diff
@@ -0,0 +1,52 @@
...
+#define SPOTFLOW_METRICS_AGGREGATOR_H_
+
+#include "spotflow_metrics_types.h"
+#include "esp_timer.h"
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/* External message queue (defined in spotflow_metrics_net.c) */
+extern struct QueueHandle_t g_spotflow_metrics_msgq;
```
roberthusak: This is not necessary, because the function `spotflow_metrics_enqueue` is used instead.

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system_connection.c
```diff
@@ -0,0 +1,51 @@
...
+
+static struct spotflow_metric_int* g_connection_state_metric;
+
+/**
+ * @brief Initialize connection metrics
+ *
+ * Registers connection_mqtt_connected metric.
+ *
+ * @return Number of metrics registered on success, negative errno on failure
+ */
```
roberthusak: Again, the comments of each function are already in the header file, remove them from here.

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_reset_helper.c
```diff
@@ -0,0 +1,101 @@
...
+	{ ESP_RST_USB, "USB" },
+	{ ESP_RST_JTAG, "DEBUG" },
+	{ ESP_RST_EFUSE, "SECURITY" },
+	{ ESP_RST_PWR_GLITCH, "POWER_GLITCH" },
+	{ ESP_RST_CPU_LOCKUP, "CPU_LOCKUP" },
+};
+
+#define RESET_CAUSE_COUNT (sizeof(reset_cause_map) / sizeof(reset_cause_map[0]))
+
+static void reset_cause_to_string(uint32_t cause, char* buf, size_t buf_len)
```
roberthusak: Static functions should be located at the end of the file, see `CODING_STYLE.md`.

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system.c
```diff
@@ -0,0 +1,164 @@
...
+#endif
+
+#ifdef CONFIG_SPOTFLOW_METRICS_SYSTEM_RESET_CAUSE
+	spotflow_report_reboot_reason();
+#endif
+
+	/* Create ESP-IDF one-shot timer */
+	const esp_timer_create_args_t timer_args = { .callback = &collection_timer_handler,
+						     .name = "system_metrics_timer" };
+	esp_timer_create(&timer_args, &g_collection_timer);
```
jmasek: not checking error but other places does.

roberthusak (commented):
I checked most of the files, will finish the review on Monday.

esp_idf/spotflow/device_sdk/examples/metrics/main/app_main.c
```diff
@@ -0,0 +1,218 @@
...
+/* ========================= */
+
+static int init_application_metrics(void)
+{
+    int rc;
+    rc = spotflow_register_metric_int("app_counter",
+                                     SPOTFLOW_AGG_INTERVAL_1MIN,
+                                     &g_app_counter_metric);
+    if (rc < 0) {
+        ESP_LOGI(TAG,"Failed to register app_counter metric: %d", rc);
```
jmasek: err?

jmasek (changes requested):
Complete review done, let me know when comments addressed or attention needed.

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system_cpu.c
noorhaq: Nope it won't.

esp_idf/spotflow/device_sdk/CMakeLists.txt
noorhaq: Some of the global lwip sets needed which were not exposed throgh Kconfig for getting network stats.

esp_idf/spotflow/device_sdk/examples/metrics/sdkconfig.defaults
```diff
@@ -0,0 +1,13 @@
+CONFIG_SPOTFLOW_DEVICE_ID="qemu"
```
roberthusak: Shouldn't this value be only in `sdkconfig.qemu` as in other samples?

esp_idf/spotflow/device_sdk/examples/metrics/main/app_main.c
```diff
@@ -0,0 +1,218 @@
...
+
+        if (iteration % 10 == 0) {
+            ESP_LOGI(TAG,"Periodic health check at iteration %d", iteration);
+        }
+
+        vTaskDelay(pdMS_TO_TICKS(2000));
+    }
+
+	int test_num = 1;
+	while (1) {
```
roberthusak, line 88: It's unnecessary to keep this loop from the logs sample, please remove it.

esp_idf/spotflow/device_sdk/examples/metrics/main/app_main.c
```diff
@@ -0,0 +1,218 @@
...
+			ESP_LOGW(TAG, "Test Number. %d", test_num++);
+			ESP_LOGI(TAG, "[APP] Free memory: %" PRIu32 " bytes",
+				 esp_get_free_heap_size());
+		}
+		vTaskDelay(50000/portTICK_PERIOD_MS); // Delay for 50s
+	}
+}
+
+
+/* ========================= */
```
roberthusak: Wouldn't it be better to keep the structured comments from the Zephyr sample instead of these "horizontal lines"?

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system_network.c
```diff
@@ -0,0 +1,178 @@
...
+ * @return int
+ */
+int spotflow_metrics_system_network_init(void)
+{
+	int rc;
+
+	memset(g_hooks, 0, sizeof(g_hooks));
+
+	/* Register netif extended callback — fires when any netif is added */
+	netif_add_ext_callback(&g_netif_callback, netif_ext_callback);
```
roberthusak, line 77: Is it guaranteed that the callback will be registered **before** the net interfaces are added? So that we don't miss any.

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system_network.c
```diff
@@ -0,0 +1,178 @@
...
+static netif_ext_callback_t g_netif_callback;
+
+/**
+ * @brief Hook for handling network transmit operations
+ *
+ * @param netif
+ * @param p
+ * @return err_t
+ */
+static err_t hooked_linkoutput(struct netif* netif, struct pbuf* p)
```
roberthusak: Static functions should be defined after public ones, see `CODING_STYLE.md`.

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system_network.c
```diff
@@ -0,0 +1,178 @@
...
+{
+	for (int i = 0; i < CONFIG_SPOTFLOW_METRICS_SYSTEM_NETWORK_MAX_INTERFACES; i++) {
+		if (g_hooks[i].active && g_hooks[i].lwip_netif == netif) {
+			g_hooks[i].tx_bytes += p->tot_len;
```
roberthusak: Assigning a 64-bit value on a 32-bit system doesn't have to be atomic (with regards to reading the value from `spotflow_metrics_system_network_collect`), right? Would it make sense to add a spinlock or something like that here?

This applies to handling received bytes as well.

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system_stack.c
```diff
@@ -0,0 +1,298 @@
...
+{
+	if (!g_stack_free_metric) {
+		SPOTFLOW_LOG("Stack metric not registered");
+		return;
+	}
+
+#ifdef CONFIG_SPOTFLOW_METRICS_SYSTEM_STACK_ALL_THREADS
+	// Enumerate all tasks
+	UBaseType_t task_count = uxTaskGetNumberOfTasks();
+	TaskStatus_t* task_array = pvPortMalloc(task_count * sizeof(TaskStatus_t));
```
roberthusak: Why not `malloc` and `free`?

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system_stack.c
roberthusak: Furthermore, is the "high water mark" the value that we really want? Currently, the result is the maximum stack size since boot, right? AFAIK we want the current stack size, which I suppose can be calculated from `pxTopOfStack`.

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system_stack.c
```diff
@@ -0,0 +1,298 @@
...
+
+#ifndef CONFIG_SPOTFLOW_METRICS_SYSTEM_STACK_ALL_THREADS
+#define CONFIG_SPOTFLOW_METRICS_SYSTEM_STACK_MAX_THREADS 4
+static TaskHandle_t g_tracked_threads[CONFIG_SPOTFLOW_METRICS_SYSTEM_STACK_MAX_THREADS];
+static SemaphoreHandle_t g_tracked_threads_mutex;
+static bool g_tracked_threads_initialized;
+#endif
+
+typedef struct
+    tskTaskControlBlock /* The old naming convention is used to prevent breaking kernel aware debuggers. */
```
roberthusak, line 29: As this structure is taken from the FreeRTOS implementation, is it guaranteed to be stable across the ESP-IDF versions we support? I.e., 5.0+?

Please also add a comment to the code explaining why it's necessary to define it here. I suppose it's because some of the pointers aren't exposed publicly.

esp_idf/spotflow/device_sdk/src/metrics/spotflow_metrics_aggregator.c
```diff
@@ -0,0 +1,346 @@
...
+					     (interval_ms - jitter) * 1000ULL);
+			ctx->timer_started = true;
+		}
+	}
+
+	xSemaphoreGive(metric->lock);
+	return 0;
+}
+
+/* --- Internal functions --- */
```
roberthusak: I'd remove this line for consistency (same with "Forward declarations" above) - we don't have it in any other file.

esp_idf/spotflow/device_sdk/src/metrics/spotflow_metrics_aggregator.c
```diff
@@ -0,0 +1,346 @@
...
+
+static void aggregation_timer_callback(void* arg)
+{
+	struct metric_aggregator_context* ctx = (struct metric_aggregator_context*)arg;
+	struct spotflow_metric_base* metric = ctx->metric;
+	int64_t timestamp_ms = esp_timer_get_time() / 1000ULL;
+
+	if (xSemaphoreTake(metric->lock, portMAX_DELAY) != pdTRUE)
+		return;
+
```
roberthusak, line 337: Shouldn't we keep the internal message "Aggregation window closed for metric ..." for debugging?

esp_idf/spotflow/device_sdk/src/metrics/spotflow_metrics_aggregator.c
```diff
@@ -0,0 +1,346 @@
...
+static void init_timeseries_aggregation_state(struct metric_timeseries_state* ts,
+					      enum spotflow_metric_type type)
+{
+	if (type == SPOTFLOW_METRIC_TYPE_INT) {
+		ts->min_int = INT64_MAX;
+		ts->max_int = INT64_MIN;
+	} else if (type == SPOTFLOW_METRIC_TYPE_FLOAT) {
+		ts->min_float = FLT_MAX;
+		ts->max_float = -FLT_MAX;
+	}
```
roberthusak, line 220: The `else` branch with logging error from the Zephyr code should be kept here - it could help, for example, with extending the enum later.

esp_idf/spotflow/device_sdk/src/metrics/spotflow_metrics_heartbeat.c
```diff
@@ -0,0 +1,138 @@
...
+	uint8_t* payload = NULL;
+	size_t len = 0;
+	int rc = spotflow_metrics_cbor_encode_heartbeat(uptime_ms, &payload, &len);
+	if (rc != 0) {
+		SPOTFLOW_LOG("Failed to encode heartbeat: %d", rc);
+		return;
+	}
+
+	/* Allocate message structure */
+	struct spotflow_mqtt_metrics_msg* new_msg = malloc(sizeof(*new_msg));
```
roberthusak: https://github.com/spotflow-io/device-sdk/pull/77 contains improved handling of heartbeat metric for Zephyr - it contains only one buffer for payload instead of allocating it dynamically. Can you incorporate this logic to the ESP-IDF module as well?

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_reset_helper.c
```diff
@@ -0,0 +1,101 @@
...
+    ESP_RST_EFUSE,      //!< Reset due to efuse error
+    ESP_RST_PWR_GLITCH, //!< Reset due to power glitch detected
+    ESP_RST_CPU_LOCKUP, //!< Reset due to CPU lock up (double exception)
+*/
+static const struct {
+	uint32_t flag;
+	const char* name;
+} reset_cause_map[] = {
+	{ ESP_RST_UNKNOWN, "UNKNOWN" },
+	{ ESP_RST_POWERON, "POWER_ON" },
```
roberthusak, line 33: @jmasek Should we map `RESET_POR` in Zephyr to `"POWER_ON"` (instead of `"POR"`) as well so that it's unified?

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_reset_helper.c
```diff
@@ -0,0 +1,101 @@
...
+	{ ESP_RST_EXT, "PIN" },
+	{ ESP_RST_SW, "SOFTWARE" },
+	{ ESP_RST_PANIC, "PANIC" },
+	{ ESP_RST_INT_WDT, "INT_WDT" },
+	{ ESP_RST_TASK_WDT, "TASK_WDT" },
```
roberthusak: Shouldn't both `ESP_RST_INT_WDT` and `ESP_RST_TASK_WDT` be mapped to `"WATCHDOG"`? Or maybe `"INTERRUPT_WATCHDOG"` and `"TASK_WATCHDOG"`?

roberthusak (commented):
Review finished. Please let me know when it's ready to be checked again.

esp_idf/spotflow/device_sdk/src/net/spotflow_mqtt.c
noorhaq: Yes. But I found a bug in the logging part where the logging K config is actually not connected. I will fix that in a separate branch and PR.

esp_idf/spotflow/device_sdk/src/metrics/spotflow_metrics_heartbeat.c
noorhaq: The g_pending_heartbeat is being freed here when it is moved to the msg [Line 103](https://github.com/spotflow-io/device-sdk/pull/73/changes/BASE..41b752c269490ae3c51ea48083b9f04f3c8f0d35#diff-fbd703a242b7a9137c5856d9bff817865cebadcfcbfd5830162eb0622bd27230R101).
So at this time it is already free. Checked with debugger too.

esp_idf/spotflow/device_sdk/examples/metrics/main/app_main.c
noorhaq: Removed.

esp_idf/spotflow/device_sdk/src/metrics/spotflow_metrics_net.c
noorhaq: Fixed.

esp_idf/spotflow/device_sdk/src/metrics/spotflow_metrics_aggregator.c
noorhaq: Fixed.

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system.c
noorhaq: Also this is unlikely to happen. i.e. it will only happen in case metrics init failed due to timer failure which will cause a reset. So it will voluntarily give up the CPU and other resources to other tasks.

esp_idf/spotflow/device_sdk/Kconfig.metrics_systems
noorhaq: Fixed.

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system_stack.c
noorhaq: Yes, I tested and checked with both 5.0 and 6.
And in future ones we can just update it with like ESP_IDF Version tag to support multiple in case it changes over versions.

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system_stack.c
noorhaq: This is thread safe as we are using multiple threads etc so this is thread safe allocation.

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system_stack.c
noorhaq: Yep I added `pxTopofStack` change. Like this just gets unused stack space to allows me to calculate the used stack space. Plus I think we should send this as well. Like showing a red line or something in alerts that at any point in system thread/task came close to stack overflow.

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system_network.c
noorhaq: Nope it is to call if any new interface is added while the code is running. For checking the previous ones we already have the `esp_netif_get_nr_of_ifs` to iterate over the previous ones. As after the initialization we aren't connecting the hooks to the new interface. As this will iterate over the new interface to `g_hooks[i].active` And disable the old one if we don't have space.

esp_idf/spotflow/device_sdk/spotflow.c
roberthusak: @noorhaq I see that you fixed the conditional compilation in the function `spotflow_metrics`, but the header inclusion is still gated just by `CONFIG_SPOTFLOW_METRICS_HEARTBEAT`. Fix that, please.

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system.c
roberthusak: @noorhaq Not sure if I understand your comments. The problem, as a I see it, is that if the function `spotflow_metrics_system_init` returns via any other `return` than the last one, it wouldn't set `g_system_metrics_init_state` to `2`. As a result, any future caller of this function would be stuck infinitely waiting for the system to be initialized (waiting for `g_system_metrics_init_state` to be `2`).

@jmasek The main question is: Do we really need this kind of synchronization for system metrics? As I understand it, this function should be called only once from `spotflow_init`, which itself should be called only once from the user code. AFAIK `spotflow_init` isn't guarded against multiple calls, so doing so would cause much more problems than doubly initialized system metrics anyway. So I suggest removing `g_system_metrics_init_state` altogether. What do you think?

Btw. this issue originates from Zephyr's `spotflow_metrics_system_init`. Because it's called only once from the processor thread which we have under our control, I suggest removing the synchronization from the Zephyr version as well.

esp_idf/spotflow/device_sdk/include/metrics/system/spotflow_metrics_system.h
roberthusak: @noorhaq This should be fine, because metric messages aren't sent right away but buffered. Therefore, they will eventually be delivered even if the connection temporarily disconnects. `spotflow_metrics_system_report_connection_state` should be called whenever the MQTT connection connects/disconnects, as in Zephyr.

esp_idf/spotflow/device_sdk/src/metrics/spotflow_metrics_net.c
roberthusak: @noorhaq It's not certain that we'll try again - non-aggregated metrics (events) are just skipped when the buffer is currently full. Also, `ENOBUFS` seems like the best description of reality - the actual problem is that the sending buffer is full, it's up to the caller to decide how to handle that (e.g., to try again later or give up). Furthermore, the return value is propagated to public functions such as `spotflow_report_event`, these already contain `ENOBUFS` in their descriptions.

esp_idf/spotflow/device_sdk/src/metrics/spotflow_metrics_net.c
roberthusak: Looks good, just please address the second paragraph from the original comment (documenting this behavior in the header).

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system_stack.c
roberthusak: I agree that that the "high water mark" (i.e., the maximum stack consumed in the current rum) metric could be useful, but I wouldn't do it as a part of this PR. It would need a separate Kconfig option (and maybe smaller sample rate) since it's AFAIK more computation intensive (`uxTaskGetStackHighWaterMark` must count all the free bytes in all stacks). Furthermore, we'd need to somehow incorporate it into the system metrics in the GUI.

I'll discuss with the rest of the team if we want to add it later. But for the sake of this PR, please remove `uxTaskGetStackHighWaterMark` and report just the current metrics (free bytes and stack usage percentage).

Btw. the computation of `stack_size` doesn't seem right. `pxTopOfStack` (according to [this comment](https://github.com/spotflow-io/device-sdk/blob/df73e6d798a903312c889ab399bdd3d98caadadf/esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system_stack.c#L31)) changes by the currently last placed item, so the right formula is AFAIK `stack_size = pxEndOfStack - pxStack`.

`pxTopOfStack` should be then used to calculate `used_bytes`. The formula for `used_bytes` will AFAIK differ by the stack growth direction - for upward growth (`portSTACK_GROWTH > 0`) `pxTopOfStack - pxStack` and for downward growth (`portSTACK_GROWTH < 0`) `pxEndOfStack - pxTopOfStack`. Correct me if I'm wrong.

esp_idf/spotflow/device_sdk/examples/metrics/main/app_main.c
roberthusak: Error indeed makes more sense here.

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system_network.c
roberthusak: I see that you added hooking the already existing interface, that's good 👍 Just two more requests:

1) Please extract the loop labeled as "Find a free slot and install hooks" to a separated function - it's quite complex and used at two places.
2) If the callback can be executed from a different thread than the function `spotflow_metrics_system_network_init` (which I suppose it can), please add a mutex (or something equivalent) to protect the modification of `g_hooks`. One mutex for the whole array should be enough.

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system_network.c
roberthusak: Or maybe use the mutex mentioned in [this comment](https://github.com/spotflow-io/device-sdk/pull/73#discussion_r3252723986). I'm not sure what's their overhead though (so that we don't throttle network communication), so maybe something more lightweight would be better.

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system_stack.c
roberthusak: According to the ESP-IDF documentation (AFAIK not in FreeRTOS in general though), `malloc` and `free` are thread-safe as well: https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/mem_alloc.html#thread-safety

Which is good, because otherwise we would need to replace all the other calls to `malloc` and `free` in other parts of the code 😅

esp_idf/spotflow/device_sdk/src/metrics/spotflow_metrics_heartbeat.c
```diff
@@ -0,0 +1,160 @@
...
+	int rc = spotflow_mqtt_publish_message(SPOTFLOW_MQTT_LOG_TOPIC, msg.payload, msg.len,
+					       SPOTFLOW_MQTT_LOG_QOS);
+
+	if (rc == -EAGAIN) {
+		/* Schedule retry */
+		if (g_retry_count < 3) {
+			g_next_retry_time_us = now_us + (retry_delays_ms[g_retry_count] * 1000);
+			g_retry_count++;
+			SPOTFLOW_DEBUG("MQTT busy, retry #%d scheduled in %d ms", g_retry_count,
+				       retry_delays_ms[g_retry_count - 1]);
```
roberthusak, line 129: These 5 lines aren't protected by the mutex. Maybe extend the one in the `else` branch?

esp_idf/spotflow/device_sdk/src/metrics/spotflow_metrics_net.c
```diff
@@ -0,0 +1,110 @@
...
+	msg->payload = payload;
+	msg->len = len;
+
+	/* Try to enqueue */
+	if (xQueueSend(g_spotflow_metrics_msgq, &msg, 0) == pdTRUE) {
+		spotflow_mqtt_notify_action(SPOTFLOW_MQTT_NOTIFY_METRICS);
+		return 0;
+	}
+
+	/* Queue is full → drop one old metirc */
```
roberthusak, line 91: @jmasek Do we want this "drop oldest" logic for metrics? If so, we should use it in the Zephyr implementation as well.

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system.c
noorhaq: Kind of correct. If the metrics are initialized then it will simply return if not like due to some initialization is not complete and again initialization is called. Then It will put the thread/task in infinite wait to free up system resources.

esp_idf/spotflow/device_sdk/src/metrics/spotflow_metrics_net.c
noorhaq: How is it being handled in zephyr currently? Like I thought this was the sequence like from logs and such otherwise we can risk overflow. Or no new entry of metrics.

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_reset_helper.c
jmasek: yes a good idea.

esp_idf/spotflow/device_sdk/src/net/spotflow_mqtt.c
jmasek: any progress on fix?

esp_idf/spotflow/device_sdk/src/metrics/spotflow_metrics_net.c
jmasek: right now, we use drop-new because there are no priorities in metrics, and also is is open for improvement - it backpressures the aggregator so it could potentially increase the aggregation window if needed. It is safer for events. For event-like metrics, older events are not necessarily less important than newer ones. A boot/reset/connection event should not be displaced by a later temperature sample.
It avoids unfairness. A high-rate metric can evict older lower-rate metrics if drop-oldest is global.

esp_idf/spotflow/device_sdk/src/net/spotflow_mqtt.c
noorhaq: It should take around a day to a day and a half. I just need to add some preprocessor configurations. I’ll merge the files once the metrics PR is merged.

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system_network.c
noorhaq: Callback will be executed from other tasks mostly netif. Like when interface is changed like ethernet to wifi or cellular and vice versa and I tried it and it didn't cause any issues. With a mutex in which only the init function calls and controls the flow of g_hooks will not let any interface added or changed while the code is running as init will run once and rest of the time the interfaces will be updated by the callback.

esp_idf/spotflow/device_sdk/src/metrics/spotflow_metrics_net.c
noorhaq: TBH we just have to add priorities in our queue logic. And then we can pass this. Like the overall artitecture can take a little bit of planning on how to handle it. Rest for both logs and metrics it will same. In my mind I have 2 ways to do it.
One instead of variable sizes in queue we have a fixed sized block like 1024 so we can just overwrite any specific msg in queue like we have 5 msgs.
1. HIGH Otherwise it will remove this one. And insert the new high prority msg here. Or keep it and just drop any new msg because it could be cascading issue.
2. HIGH
3. LOW  it will only remove and insert the new msg here. Like the sequence number will decided kind of which is the latest msg.
4. MEDIUM
5. HIGH

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system.c
noorhaq: @jmasek What should we do here.

esp_idf/spotflow/device_sdk/src/metrics/spotflow_metrics_net.c
jmasek: lets keep current strategy.

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system_network.c
roberthusak: Nevermind, one `g_hooks[i]` will be accessed only from one place at a time, so no synchronization is necessary.

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system.c
noorhaq: WE will do it in a separate PR. To keep it consistent with the zephyr code.

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system_cpu.c
jmasek: What won't explain? Please explain.

noorhaq: The timer callback
Like what it does is, checks the idle time for cpu core at first. Like to get the baseline. 
Than waits for 1s and then again measure that base line. So it doesn't contradict with anything. Or interfere with any timer functionality. As all these are running in the background and we are just getting those values at set intervals.

noorhaq: If you think the delay here can interfere with the timer callback then it won't happen as well.
As this is task delay the timer is running in the background so those ticks are being monitored. And this is just pausing just this task/function for 1s.

esp_idf/spotflow/device_sdk/src/metrics/spotflow_metrics_aggregator.c
jmasek: The ownership discussion looks incomplete for error paths. flush_no_aggregation_metric() still frees the payload after enqueue failure, while spotflow_metrics_enqueue() also frees it on queue-full retry failure (resulting in double-free); the aggregated path (lines 318-320) has the opposite problem and can leak on enqueue failure. Please reopen this and make the ownership contract consistent for both success and failure cases.

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system.c
jmasek: legit, it shouldn't be called multiple times. Therefore, I suggest removing synchronization logic. We should not keep the invalid code there. If we are not removing it, failure should reset the state back to 0.

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system_cpu.c
jmasek: yes, but the documentation says 

> It is recommended to keep callbacks as short as possible to avoid delaying other callbacks.

https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/esp_timer.html#general-procedure

Also documentation states 

> Dispatches timer callbacks from a single high-priority ESP Timer task (esp_timer task (notified by ISR) > callback).

It means that all the timer tasks are dispatched from the same task -> meaning it delays also other timers that uses task dispatch - eg. heartbeat timer and possibly other user's timers

esp_idf/spotflow/device_sdk/src/net/spotflow_mqtt.c
jmasek: please add the PR link to this issue so it can be tracked independently.

jmasek:
New issue: sp_idf/spotflow/device_sdk/include/metrics/spotflow_metrics_cbor.h:18-20 still says callers should free the returned buffer with k_free(), but this ESP-IDF implementation allocates with malloc() / ree() in src/metrics/spotflow_metrics_cbor.c. The header should say ree() here, otherwise it documents the wrong deallocator for callers.

esp_idf/spotflow/device_sdk/spotflow.c
```diff
@@ -17,10 +17,36 @@
...
+#ifdef CONFIG_SPOTFLOW_METRICS
+void spotflow_metrics(void)
+{
+	spotflow_metrics_init();
+	spotflow_metrics_net_init();
+#ifdef CONFIG_SPOTFLOW_METRICS_HEARTBEAT
+	spotflow_metrics_heartbeat_init();
+#endif
+#ifdef CONFIG_SPOTFLOW_METRICS_SYSTEM
+	spotflow_metrics_system_init();
```
jmasek, line 43: Potential issue: the initial MQTT connection-state event may be lost due to initialization order. In spotflow_init(), MQTT is started before metrics initialization, so if MQTT_EVENT_CONNECTED fires early, spotflow_metrics_system_report_connection_state(true) can return before system metrics are initialized and the initial connected event is never retried. This may or may not be acceptable depending on the intended semantics of the connection-state metric, but I think it should be called out explicitly and discussed. @roberthusak

esp_idf/spotflow/device_sdk/Kconfig.projbuild
```diff
@@ -1,5 +1,9 @@
 menu "Spotflow"
-
+    config SPOTFLOW_MQTT_TASK_PRIORITY
+        int "Task priority for MQTT tasks to send logs, coredump and other logs data to cloud"
```
jmasek, line 3: `CONFIG_SPOTFLOW_MQTT_TASK_PRIORITY` should be wired into `xTaskCreate()` instead of leaving the MQTT publish task at a hardcoded priority `5`. Right now the Kconfig option does not affect runtime behavior. Also, the help text here says "Lower number means higher priority", which is backwards for FreeRTOS task priorities. Please wire the config into task creation and fix the help text accordingly.

esp_idf/spotflow/device_sdk/src/metrics/spotflow_metrics_aggregator.c
```diff
@@ -0,0 +1,353 @@
...
+	uint64_t seq_num = metric->sequence_number++;
+
+	int rc = spotflow_metrics_cbor_encode_aggregated(metric, ts, timestamp_ms, seq_num,
+							 &cbor_data, &cbor_len);
+	if (rc < 0) {
+		reset_timeseries_state(metric, ts);
+		return rc;
+	}
+
+	rc = spotflow_metrics_enqueue(cbor_data, cbor_len);
```
jmasek, line 318: In `flush_timeseries()`, the timeseries is reset after CBOR encode failure and after successful enqueue, but not after enqueue failure. That means if the aggregation window closes and enqueue fails, the old samples stay in the timeseries and can be included again in a later flush, which changes the aggregation-window semantics and can duplicate counts/sums/min/max into the next interval. Zephyr resets the timeseries on both success and failure here; I think ESP-IDF should do the same.

esp_idf/spotflow/device_sdk/spotflow.c
```diff
@@ -36,4 +62,7 @@ void spotflow_init(void)
 	}
 #endif
 	spotflow_config_init();
-}
\ No newline at end of file
+#ifdef CONFIG_SPOTFLOW_METRICS
```
jmasek: There is an init-order race around the MQTT connection-state metric. `spotflow_init()` starts MQTT before metrics initialization, so if `MQTT_EVENT_CONNECTED` fires quickly, `spotflow_metrics_system_report_connection_state(true)` can return early because system metrics are not initialized yet. In that case the initial connected event is silently lost and never retried. Please decide whether that is acceptable for the intended metric semantics, or initialize/report the state in a way that preserves the initial event.

esp_idf/spotflow/device_sdk/include/metrics/spotflow_metrics_cbor.h
```diff
@@ -0,0 +1,69 @@
...
+#endif
+
+/* Current heartbeat payload encodes to ~35 bytes; keep margin for future fields. */
+#define SPOTFLOW_METRICS_HEARTBEAT_CBOR_MAX_LEN 64
+
+/**
+ * @brief Encode metric message to CBOR format
+ *
+ * Allocates and returns a CBOR-encoded buffer. Caller owns the buffer
+ * and is responsible for freeing it with k_free().
```
jmasek: still says callers should free the returned buffer with k_free(), but this ESP-IDF implementation allocates with malloc() / free()

esp_idf/spotflow/device_sdk/spotflow.c
roberthusak: That's a good point. I think we should initialize metrics before MQTT. We do that in Zephyr as well.

esp_idf/spotflow/device_sdk/spotflow.c
roberthusak: Duplicate of https://github.com/spotflow-io/device-sdk/pull/73#discussion_r3311453072?

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system_cpu.c
noorhaq: Yes that's on the callback that's why for the metrics callback we reuse the timer like create a timer once(one shot timer not a periodic one) and then recreate when the collection is complete.
First created inside the `spotflow_metrics_system_ini()` function inside the `esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system.c`
[line 109](https://github.com/spotflow-io/device-sdk/pull/73/changes/BASE..41b752c269490ae3c51ea48083b9f04f3c8f0d35#diff-67df17d2193b25532dd5d0f2e5c55a720845a5743da08383df87d4adce426d9cR110)


And then again inside the `collection_timer_handler` function same file. [Line 155](https://github.com/spotflow-io/device-sdk/pull/73/changes/BASE..41b752c269490ae3c51ea48083b9f04f3c8f0d35#diff-67df17d2193b25532dd5d0f2e5c55a720845a5743da08383df87d4adce426d9cR155)

We only created the periodic timer which can cause this issue for the heartbeat in the metrics as that has a short callback time.

esp_idf/spotflow/device_sdk/spotflow.c
noorhaq: Fixed.

esp_idf/spotflow/device_sdk/src/metrics/spotflow_metrics_aggregator.c
noorhaq: Fixed.

esp_idf/spotflow/device_sdk/src/metrics/spotflow_metrics_aggregator.c
noorhaq: Fixed.

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system.c
noorhaq: Fixed.

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system_cpu.c
noorhaq: Changed the logic.

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system_stack.c
noorhaq: Fixed.

esp_idf/spotflow/device_sdk/spotflow.c
noorhaq: Fixed.

esp_idf/spotflow/device_sdk/Kconfig.projbuild
jmasek: The priority is wired now, thanks. The help text still needs one small fix: FreeRTOS uses higher numeric values for higher priority, so "Lower number means higher priority" is backwards. Please change it to say that a higher number means higher priority.

esp_idf/spotflow/device_sdk/include/metrics/spotflow_metrics_net.h
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * Dequeues one message from the metrics queue and publishes via MQTT.
+ * Called repeatedly by processor thread to drain the queue.
+ *
+ * @return 1 if message processed, 0 if queue empty, negative errno on failure
+ */
+int spotflow_poll_and_process_enqueued_metrics(void);
+
+/**
+ * @brief Enqueue a metric message for sending, by conforming it to
```
jmasek, line 29: Can we update this function comment to match the current ownership and return-value contract? `spotflow_metrics_enqueue()` takes ownership of `payload` once called, including failure paths where the current message can be freed after queue-full retry failure, so callers must not free `payload` after calling it. The return docs also look stale: the implementation returns `0` on success and negative errno on failure, not `1` on success / `0` if full.

esp_idf/spotflow/device_sdk/Kconfig.projbuild
noorhaq: Fixed.

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system_network.c
noorhaq: @roberthusak Fixed it. Like mutex not needed here as it's execution is controlled.

esp_idf/spotflow/device_sdk/src/net/spotflow_mqtt.c
noorhaq: Yes Added PR Link
 https://github.com/spotflow-io/device-sdk/pull/79

esp_idf/spotflow/device_sdk/src/metrics/system/spotflow_metrics_system.c
```diff
@@ -102,14 +105,33 @@ int spotflow_metrics_system_init(void)
...
 }
 
+static void collection_timer_handler(void* arg)
+{
+	BaseType_t ret;
+
+	/* Spawn worker task */
+	ret = xTaskCreate(spotflow_metrics_collection_task, "spotflow_metrics_task",
+			  4096, /* stack size */
+			  NULL, tskIDLE_PRIORITY + 1, NULL);
```
roberthusak, line 115: This solution allocates and runs a new short-lived thread (task) for each system metrics collection, right? Wouldn't it be better to instead have a static thread that wakes up periodically?

noorhaq: Nope if the duration of metrics was short then yes but since it can increase quite by a margin it is better to collect the metrics and send/upload it to queue to upload to cloud and delete the task to free up its resources. Like say if we only collect metrics every day or every hour then a sleeping task has way too much overhead.

esp_idf/spotflow/device_sdk/include/metrics/spotflow_metrics_net.h
noorhaq: Done

roberthusak (dismissed):

jmasek (approved):

roberthusak (approved):

PR #78, opened 2026-05-29
Fix/coredump crashloop

jmasek:


noorhaq:
These are mainly changes in the logging part not sure how it helped the coredump.

jmasek:
> These are mainly changes in the logging part not sure how it helped the coredump.

yes because the coredump example was crashing.

roberthusak (approved):

PR #82, opened 2026-06-16
BLE 

jmasek:


zephyr/src/coredumps/spotflow_coredumps_backend.c
```diff
@@ -43,14 +44,19 @@ struct coredump_info {
...
-static int enqueue_log_msg(const struct spotflow_mqtt_coredumps_msg* msg)
+static int enqueue_log_msg(const struct spotflow_coredump_msg* msg)
 {
 	int rc = k_msgq_put(&g_spotflow_core_dumps_msgq, &msg, K_FOREVER);
 	return rc;
 }
 
 static void spotflow_coredumps_thread_entry(void)
 {
+	if (!spotflow_transport_supports_feature(SPOTFLOW_TRANSPORT_FEATURE_COREDUMPS)) {
```
jmasek: temporary - before implementing the coredump support in ble

.github/workflows/zephyr.yml
```diff
@@ -49,7 +79,9 @@ jobs:
           ${{ github.event.inputs.build_logs_all_boards && '--build-logs-for-all-boards' || '' }}
 
   build:
-    needs: generate-matrix
+    needs:
+      - unit-tests
```
roberthusak:
```suggestion
      # - unit-tests
```

See https://github.com/spotflow-io/device-sdk/actions/runs/27939169199.

zephyr/samples/ble/BLE Design POC.md
```diff
@@ -0,0 +1,152 @@
...
+
+### Connection Establishment
+
+1. Scan for and pair a device with Spotflow GATT service advertised
+2. Read Capabilities Characteristic → check version compatibility
+3. Initial Information Read
+    1. Read `Device Id`
+    2. Read `Session Metadata`
+4. Establish a new MQTT Connection with the obtained device id and gateway-provided ingest-key
+    1. Connect to `ingest-cbor` , `config-cbor-d2c` for writes and `config-cbor-c2d` for reads
```
roberthusak: Nitpick:

```suggestion
    1. Subscribe to `config-cbor-c2d`
```

There's no "connect for reads/writes" in MQTT, only "publish" and "subscribe".

zephyr/samples/ble/BLE Design POC.md
roberthusak, line 1: Will these Markdown files be kept here after finishing the PoC? Or will we move their content into public docs? If we want to keep them here, I suggest moving them to `zephyr/docs` (or similar). They describe the feature itself rather than the sample.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 ### Added
 * Added Metrics in ESP IDF
+* Added initial Zephyr BLE log transport for Spotflow log backend, with a GATT metadata service and CBOR log notifications.
```
roberthusak: Metrics as well, right?

zephyr/samples/ble/BLE Design POC.md
jmasek: no not keeping it here, it will be removed before merging - it is poc design.

CHANGELOG.md
jmasek: changed

zephyr/samples/ble/sample.yaml
roberthusak, line 1: Do we need this file? It's not used in other samples.

zephyr/src/net/Kconfig.ble
```diff
@@ -0,0 +1,38 @@
...
+	int "Maximum supported BLE ATT MTU"
+	default 247
+	range 23 247
+	help
+		Maximum ATT MTU this device supports for the Spotflow BLE transport.
+		The negotiated ATT MTU for a connection may still be lower, depending on
+		the central capabilities and whether the central initiates MTU exchange.
+
+configdefault BT_L2CAP_TX_MTU
+	default SPOTFLOW_BLE_ATT_MTU
```
roberthusak: Do we really need `SPOTFLOW_BLE_ATT_MTU` as a separate option? Wouldn't it be simpler to remove it and just suggest `247` as the default value of `BT_L2CAP_TX_MTU`? I.e., do it the same way as for `BT_BUF_ACL_TX_SIZE` and `BT_BUF_ACL_RX_SIZE`?

zephyr/src/coredumps/spotflow_coredumps_backend.c
```diff
@@ -43,14 +44,19 @@ struct coredump_info {
 static struct coredump_info coredump_info;
 
 /* blocks forever until space is available */
-static int enqueue_log_msg(const struct spotflow_mqtt_coredumps_msg* msg)
+static int enqueue_log_msg(const struct spotflow_coredump_msg* msg)
```
roberthusak: Can you rename the function to `enqueue_coredump_msg` or similar? I know it was named wrong even before the PR, but since we're modifying this file anyway..

zephyr/src/net/spotflow_processor.c
```diff
@@ -35,55 +42,73 @@
...
 
-	LOG_DBG("Spotflow registered TLS credentials");
 #ifdef CONFIG_SPOTFLOW_METRICS
 	spotflow_metrics_net_init();
 #ifdef CONFIG_SPOTFLOW_METRICS_HEARTBEAT
 	spotflow_metrics_heartbeat_init();
 #endif
 #endif
 
+#if CONFIG_SPOTFLOW_TRANSPORT_MQTT
```
roberthusak, line 78: Would it make sense to you to eventually move MQTT-specific implementation from `spotflow_processor.c`, e.g., to `spotflow_mqtt_transport.c`? For the PoC it's IMO OK as it is though.

roberthusak: If we decide to do so, I also suggest removing `spotflow_transport.c‎` and implementing the functions declared in `spotflow_transport.h` directly in `spotflow_ble_transport.c‎` and `spotflow_mqtt_transport.c` (or `spotflow_transport_ble.c‎` and `spotflow_transport_mqtt.c‎` - this naming pattern seems more common).

zephyr/src/net/CMakeLists.txt
```diff
@@ -1,19 +1,27 @@
...
-        spotflow_connection_helper.c
         spotflow_device_id.c
-        spotflow_mqtt.c
         spotflow_session_metadata.c
-        spotflow_tls.c
+        spotflow_transport.c
 )
 
-zephyr_library_link_libraries(mbedTLS)
+if(CONFIG_SPOTFLOW_LOG_BACKEND AND CONFIG_SPOTFLOW_TRANSPORT_MQTT)
```
roberthusak: Do I understand correctly that the MQTT transport implementation now requires `CONFIG_SPOTFLOW_LOG_BACKEND` to be enabled, while the BLE transport implementation does not?

zephyr/src/net/CMakeLists.txt
```diff
@@ -1,19 +1,27 @@
...
+        spotflow_ble_transport.c
+        spotflow_ble_gatt.c
+        spotflow_ble_framing.c
+)
 
-if (CONFIG_SPOTFLOW_LOG_BACKEND)
-    if (NOT CONFIG_SPOTFLOW_DEVICE_ID)
-        message(FATAL_ERROR  "CONFIG_SPOTFLOW_DEVICE_ID must be specified
-                              if CONFIG_SPOTFLOW_LOG_BACKEND enabled")
-    endif ()
```
roberthusak, line 16: Shouldn't be `CONFIG_SPOTFLOW_DEVICE_ID` checked for both transport methods (MQTT and BLE)?

zephyr/samples/ble/sample.yaml
jmasek: no, removed

zephyr/src/coredumps/spotflow_coredumps_backend.c
jmasek: omg, good point :D, renamed

zephyr/src/net/CMakeLists.txt
jmasek: there is quite a complex logix so I am not sure it is really needed
1. If application provides spotflow_override_device_id(), use that.
2. Else if CONFIG_SPOTFLOW_DEVICE_ID is non-empty, use that.
3. Else generate one from Zephyr hardware info via hwinfo_get_device_id().
4. If hwinfo_get_device_id() fails, fall back to "default_device_id".

roberthusak: OK, makes sense 👍

zephyr/src/net/transport/ble/spotflow_ble_transport_internal.h
```diff
@@ -0,0 +1,109 @@
...
+#define SPOTFLOW_BLE_MAX_METRIC_PAYLOAD_LEN CONFIG_SPOTFLOW_METRICS_CBOR_BUFFER_SIZE
+#else
+#define SPOTFLOW_BLE_MAX_METRIC_PAYLOAD_LEN 0
+#endif
+
+#if SPOTFLOW_BLE_MAX_METRIC_PAYLOAD_LEN > SPOTFLOW_BLE_MAX_LOG_PAYLOAD_LEN
+#define SPOTFLOW_BLE_MAX_INGEST_PAYLOAD_LEN SPOTFLOW_BLE_MAX_METRIC_PAYLOAD_LEN
+#else
+#define SPOTFLOW_BLE_MAX_INGEST_PAYLOAD_LEN SPOTFLOW_BLE_MAX_LOG_PAYLOAD_LEN
+#endif
```
roberthusak: Zephyr's [`MAX` macro](https://docs.zephyrproject.org/latest/doxygen/html/group__sys-util.html#gafa99ec4acc4ecb2dc3c2d05da15d0e3f) could be useful here.

zephyr/src/net/transport/ble/spotflow_ble_transport_internal.h
```diff
@@ -0,0 +1,109 @@
...
+/* With the minimum ATT MTU of 23, the first fragment carries 15 bytes and each
+ * continuation fragment carries 17 bytes. Size the temporary frame array to the
+ * worst-case number of fragments needed for the largest configured ingest payload.
+ */
+#define SPOTFLOW_BLE_MAX_ENCODED_FRAMES                                                      \
+	(1 +                                                                                 \
+	 ((SPOTFLOW_BLE_MAX_INGEST_PAYLOAD_LEN - SPOTFLOW_BLE_FIRST_FRAGMENT_DATA_CAPACITY + \
+	   SPOTFLOW_BLE_CONT_FRAGMENT_DATA_CAPACITY - 1) /                                   \
+	  SPOTFLOW_BLE_CONT_FRAGMENT_DATA_CAPACITY))
+#endif
```
roberthusak: I recommend using [`DIV_ROUND_UP`](https://docs.zephyrproject.org/latest/doxygen/html/group__sys-util.html#gae664e7492e37d324831caf2321ddda37) for better clarity.

zephyr/src/net/transport/ble/spotflow_ble_gatt.c
```diff
@@ -0,0 +1,301 @@
...
+{
+	if (!g_spotflow_ble_transport_state.initialized) {
+		k_mutex_init(&g_spotflow_ble_transport_state.lock);
+		k_work_init_delayable(&g_spotflow_ble_transport_state.restart_advertising_work,
+				      restart_advertising_work_handler);
+		g_spotflow_ble_transport_state.initialized = true;
+	}
+
+	if (!g_spotflow_ble_transport_state.bluetooth_enabled) {
+		int rc = bt_enable(NULL);
```
roberthusak: There's a slight risk that the HCI device won't be ready yet when this function is called (similarly as the Wi-Fi driver in our samples), leading to the error `-ENODEV` and basically blocking the device from connecting until reboot. It would be safer to add at least some limited number of retries if `-ENODEV` is returned (it seems as the only error that might be transient).

zephyr/src/net/transport/ble/spotflow_ble_gatt.c
```diff
@@ -0,0 +1,301 @@
...
+		if (rc != 0 && rc != -EALREADY) {
+			LOG_ERR("Failed to enable Bluetooth: %d", rc);
+			return rc;
+		}
+
+		g_spotflow_ble_transport_state.bluetooth_enabled = true;
+		LOG_INF("Bluetooth enabled for Spotflow BLE transport");
+	}
+
+	return start_advertising();
```
roberthusak: While an advertising error from `restart_advertising_work_handler` (that's scheduled only as a reaction to disconnection) results into retrying, an error from this call leads into blocking the device from connecting until reboot. That isn't intentional, right?

zephyr/tests/ble_framing/src/main.c
```diff
@@ -0,0 +1,115 @@
...
+	size_t frame_count = 0;
+
+	int rc = spotflow_ble_transport_encode_frames(SPOTFLOW_MSG_TELEMETRY, 0x7A, payload,
+						      sizeof(payload), 20, frames,
+						      ARRAY_SIZE(frames), &frame_count);
+
+	zassert_equal(rc, 0, "unexpected rc %d", rc);
+	zassert_equal(frame_count, 1, "unexpected frame count %u", frame_count);
+	zassert_equal(frames[0].len, 9, "unexpected frame len %u", frames[0].len);
+	zassert_equal(frames[0].data[0], SPOTFLOW_MSG_TELEMETRY, NULL);
```
roberthusak: `NULL` can be omitted in all the calls to `zassert_equal` to keep the default message (`#a " not equal to " #b`), see the [Zephyr docs](https://docs.zephyrproject.org/latest/doxygen/html/group__ztest__assert.html#gacd075c7ee4dc2d64701bd3098a31b673).

zephyr/src/net/transport/ble/spotflow_ble_framing.c
```diff
@@ -0,0 +1,301 @@
...
+
+static void reset_config_rx_state_locked(void)
+{
+	g_spotflow_ble_transport_state.config_rx.active = false;
+	g_spotflow_ble_transport_state.config_rx.sequence = 0;
+	g_spotflow_ble_transport_state.config_rx.total_len = 0;
+	g_spotflow_ble_transport_state.config_rx.received_len = 0;
+}
+
+int spotflow_ble_transport_process_config_rx_frame(const void* buf, uint16_t len, uint8_t flags)
```
roberthusak, line 46: This function is quite complex, I think it would benefit from:

1) Test coverage - separate tests and/or combined encode/decode test with `spotflow_ble_transport_encode_frames`.
2) Reorganization - it would make sense to handle just the header separately (distinguishing different sizes for `is_first` of `true`/`false`) and then handle the body the same way.

zephyr/src/net/transport/ble/spotflow_ble_framing.c
```diff
@@ -0,0 +1,301 @@
...
+
+	if (len > UINT16_MAX) {
+		bt_conn_unref(conn);
+		return -EMSGSIZE;
+	}
+
+	uint16_t mtu = bt_gatt_get_mtu(conn);
+	size_t notify_payload_max = mtu > 3 ? mtu - 3 : 0;
+
+	if (notify_payload_max <= 5) {
```
roberthusak: Not sure what these numbers mean and why `3` is subtracted from `notify_payload_max`. Seems like the header size, but that should be included in the payload size, right?

Furthermore, `notify_payload_max <= 5` case is already checked in `spotflow_ble_transport_encode_frames`.

.github/workflows/zephyr.yml
```diff
@@ -29,6 +29,36 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  # will be fixed later
```
roberthusak: Discussed offline: [This job from OTA PR](https://github.com/spotflow-io/device-sdk/pull/66/changes#diff-4703d5b5938461fea21f3e87ac7f9a92daae21353c0dd41ff23696e8be8b454e) should work. We can either add it to this PR or keep the tests just local for now. Anyway, I'd get rid of the `run_twister.*` scripts.

zephyr/src/net/transport/ble/spotflow_ble_gatt.c
roberthusak: According to the description of `bt_enable`, we should also call `settings_load_subtree("bt")` when `CONFIG_BT_SETTINGS` is enabled.

zephyr/Kconfig
```diff
@@ -87,30 +51,13 @@ configdefault MBEDTLS_HEAP_SIZE
...
-
-config SPOTFLOW_INGEST_KEY
-	string "Password device will use to connect to Spotflow observability platform"
-	help
-		Ingest key is a secret key used to authenticate your device
-		It can be found in Spotflow observability platform
-
 config SPOTFLOW_MODULE_DEFAULT_LOG_LEVEL
 	int "Spotflow SDK backend level"
 	default 2
```
roberthusak: We should use `LOG_DEFAULT_LEVEL` as the default value.

zephyr/samples/ble/BLE Design POC.md
jmasek: removed

zephyr/src/net/Kconfig.ble
jmasek: what do you mean suggest - you mean tell user? like adding it to the example kconfig instead of default value here?

roberthusak: No, I thought of removing `config SPOTFLOW_BLE_ATT_MTU` and just putting `default 247` beneath `configdefault BT_L2CAP_TX_MTU`.

Because AFAIK (as I couldn't find `CONFIG_SPOTFLOW_BLE_ATT_MTU` in the code) the only job of `SPOTFLOW_BLE_ATT_MTU` is to set the default value of `BT_L2CAP_TX_MTU`. Or does it have any other purpose? The range check (`[23, 247]`) of `SPOTFLOW_BLE_ATT_MTU` doesn't prevent the user from setting `BT_L2CAP_TX_MTU` higher anyway.

roberthusak: Btw. providing a `configdefault` doesn't necessary guarantee that the provided values will be used. If we need to be sure, we should add a CMake warning/error when the values are not as expected.

zephyr/src/net/spotflow_processor.c
jmasek: done https://github.com/spotflow-io/device-sdk/pull/82/commits/5344cd5c05fd76748b1e7dd79c9e5bc2396a53b6

.github/workflows/zephyr.yml
jmasek: remove run_twister. Tests will be added in separate PR

zephyr/Kconfig
jmasek: changed

zephyr/src/coredumps/spotflow_coredumps_backend.c
jmasek: removed

zephyr/src/net/CMakeLists.txt
jmasek: removed

zephyr/src/net/transport/ble/spotflow_ble_transport_internal.h
jmasek: removed

zephyr/src/net/transport/ble/spotflow_ble_transport_internal.h
jmasek: removed

jmasek: removed

zephyr/src/net/transport/ble/spotflow_ble_framing.c
jmasek: reworked

zephyr/src/net/transport/ble/spotflow_ble_framing.c
jmasek: reworked and add more sensible names

jmasek: please check

zephyr/src/net/transport/ble/spotflow_ble_gatt.c
jmasek: fixed

zephyr/src/net/transport/ble/spotflow_ble_gatt.c
jmasek: good point, fixed

zephyr/tests/ble_framing/src/main.c
jmasek: good point, changed

zephyr/src/net/Kconfig.ble
jmasek: lgtm, changed

zephyr/tests/ble_framing/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+# SPDX-License-Identifier: Apache-2.0
+
+cmake_minimum_required(VERSION 3.20.0)
+
+find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
+project(spotflow_ble_framing_tests)
+
```
roberthusak, line 7: Adding the explicit include path should fix the issue in CI:

```suggestion

target_include_directories(app PRIVATE
        ${CMAKE_CURRENT_LIST_DIR}/../../src
)

```

.github/workflows/zephyr.yml
```diff
@@ -49,7 +132,9 @@ jobs:
           ${{ github.event.inputs.build_logs_all_boards && '--build-logs-for-all-boards' || '' }}
 
   build:
-    needs: generate-matrix
+    needs:
+      # - unit-tests
```
roberthusak: Thinking about it, should there be a dependency at all? Building of a sample and passing of unit tests seem as two complementary things.

I also suggest moving the `unit-tests` job after the `build` job as that's a common order.

zephyr/tests/ble_framing/src/main.c
roberthusak: Good, can you do it for other asserts (`zassert_mem_equal`, `zassert_true`, `zassert_false`) as well?

zephyr/src/net/spotflow_transport.h
```diff
@@ -0,0 +1,32 @@
...
+#include <stddef.h>
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef void (*spotflow_transport_message_cb)(uint8_t* payload, size_t len);
+
+enum spotflow_transport_feature {
```
roberthusak: We can now remove this whole enum as well, right?

jmasek: yes, good point, we can

.github/workflows/zephyr.yml
jmasek: definitely no

jmasek: moved

zephyr/src/net/spotflow_transport.h
jmasek: removed

zephyr/tests/ble_framing/src/main.c
jmasek: oh sorry, missed that, removed

roberthusak (approved):

PR #83, opened 2026-06-26
Feature/ble coredumps

jmasek:


.gitignore
```diff
@@ -17,3 +17,4 @@ vscode-workspaces/**
 AGENTS.md
 CLAUDE.md
 **/sdkconfig.local
+**/build-rw612
```
roberthusak: Nitpick: I suggest removing this line and using paths like `<sample>/build/<board>` for development instead.

zephyr/src/net/spotflow_ble_transport.c
```diff
@@ -22,7 +22,8 @@ bool spotflow_ble_transport_supports_feature(enum spotflow_transport_feature fea
 {
 	return feature == SPOTFLOW_TRANSPORT_FEATURE_LOGS ||
 		feature == SPOTFLOW_TRANSPORT_FEATURE_CONFIG ||
-		feature == SPOTFLOW_TRANSPORT_FEATURE_METRICS;
+		feature == SPOTFLOW_TRANSPORT_FEATURE_METRICS ||
+		feature == SPOTFLOW_TRANSPORT_FEATURE_COREDUMPS;
```
roberthusak: Since it always returns true, can we now remove this function completely, then? When adding OTA updates, we can check in build time that MQTT is used with them.

zephyr/samples/ble/src/main.c
```diff
@@ -1,49 +1,178 @@
...
+static bool was_button_pressed;
+#else
+static struct gpio_callback button_cb_data;
+#endif
+
+static void button_pressed(void)
+{
+	LOG_INF("Button pressed. Going to oops.");
+	k_oops();
+}
```
roberthusak, line 36: I suggest moving the button polling/callback-handling logic to another file and keeping just this function here. Even polling can be done in a separate thread so that the main logging loop can stay simple.

jmasek: removing, as discussed, it works on 4.4.1,

zephyr/src/net/spotflow_ble_transport.c
jmasek: yes

.gitignore
jmasek: good point, removed

jmasek:
addedd to https://github.com/spotflow-io/device-sdk/pull/82, using https://github.com/spotflow-io/device-sdk/pull/82/commits/e5e73b91a068d1045782f55ca826d1fbba3ff37e

PR #90, opened 2026-08-11
added session metadata

jmasek:


zephyr/include/spotflow/session_metadata.h
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * @brief Return labels included in each Spotflow Session Metadata message.
+ *
+ * Override this callback in application code to provide session labels. Return
+ * an empty structure when no labels should be emitted. The callback runs while
+ * the SDK encodes session metadata and must not block or call Spotflow APIs.
+ * The returned array, label keys, and string values must remain valid until
+ * the SDK has finished encoding the Session Metadata message. Keep the returned
+ * labels stable between invocations unless a new session should intentionally
+ * use different labels.
```
roberthusak: The application code doesn't know the exact time when the SDK finishes encoding the message, so I don't think the validity description is useful. AFAIK the only safe time to modify labels (when pointer to them was already returned) is during the callback itself, so maybe we should mention that?

zephyr/tests/ota/facade/src/main.c
roberthusak, line 1: I think we should create new tests for session metadata instead of adding them to the tests of OTA interface.

jmasek: lgtm.

zephyr/include/spotflow/session_metadata.h
jmasek: changed

zephyr/tests/ota/facade/src/main.c
jmasek: moved

roberthusak (approved):

